PR #5, opened 2017-01-31
Add euphoria-examples

xitep:
@d-spirov reminded me ... we're missing an examples module; so here it is :)


vanekjar:
LGTM. Thanks.

PR #7, opened 2017-01-31
#! [euphoria-spark] Upgrade to Spark 2.1.0

vanekjar:
No reason to wait with upgrade to the newest Spark. Tried running it with our benchmarks and it worked pretty well.


xitep:
👍  many thanks!

PR #17, opened 2017-02-03
Replace Dataset#getPartitioning() with Dataset#getNumPartitions()

### LINKED ISSUE #16: Redesign Dataset#getPartitioning()
The method `Dataset#getPartitioning()` in the client API doesn't make sense in its current form (spotted as part of #15). Partitioners themselves are a property of operators not datasets. However, we still need to preserve the "numPartitions" property available through the `#getPartitioning()` which is used to relay an dataset's physical split further down a flow.

I suggest to change the method to a plain `int Dataset#getNumPartitions()`. A negative value denotes "undefined" just as it is handled so far.

xitep:
* Forwarding the partitioner down the operator chain is better done through an explicit analysis of the DAG. The current approach is not type-safe. Forwarding the "parallelism", ie. partition count, though, is fine.
* Resolves #16


euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Distinct.java
```diff
@@ -71,7 +71,7 @@
         UnaryFunction<IN, ELEM> mapper /* optional */) {
```
je-ik: The following comment has nothing to do with this review, but just a suggestion to start discussion, could we replace these "/* optional */" comments with @Nullable annotation? That seems maybe a little more useful.

xitep: good point!

xitep:
thanks @je-ik

PR #26, opened 2017-02-17
Use FindBugs for static code analysis during build

### LINKED ISSUE #18: Consider documenting nullability
So far we've followed the convention that unless otherwise stated null values are not allowed. Right now, we document allowed nullability in comments. Consider switching to a stronger and more visible form of documentation by using `@javax.annotation.Nullable` and friends from the `jsr305.jar`. (We'll add add it as an explicit compile dependency even though we do inherit it transitively through the dependency on guava.)

vanekjar:
I added FindBugs check to compile phase. It fails when bug with rank `14` and lower is found (these bugs are considered as `troubling`, ` scary`, `scariest` categories).

I also added `@Nullable` annotation when explicit check for `null` values is needed.

fixes #18

### DIFF
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/util/URIParams.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/util/URIParams.java
@@ -429,7 +429,7 @@ public class URIParams {
     String value = remove(name);
     if (value == null) {
       throw new IllegalArgumentException(String.format(
-          "Dropping parameter \"{}\" which does not exist.", new Object[] { name }));
+          "Dropping parameter \"%s\" which does not exist.", name));
     }
   }
 
--- a/pom.xml
+++ b/pom.xml
@@ -402,6 +402,28 @@
         </executions>
       </plugin>
 
+      <!-- static analysis of the code using FindBugs -->
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>findbugs-maven-plugin</artifactId>
+        <version>3.0.4</version>
+        <configuration>
+          <effort>max</effort>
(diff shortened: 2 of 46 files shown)


euphoria-core/src/main/java/cz/seznam/euphoria/core/util/URIParams.java
```diff
@@ -429,7 +429,7 @@ public void dropParam(String name) {
     String value = remove(name);
     if (value == null) {
       throw new IllegalArgumentException(String.format(
-          "Dropping parameter \"{}\" which does not exist.", new Object[] { name }));
+          "Dropping parameter \"%s\" which does not exist.", name));
```
xitep: cool! 👍

pom.xml
```diff
@@ -509,6 +509,12 @@
       </dependency>
 
       <dependency>
+        <groupId>com.google.code.findbugs</groupId>
+        <artifactId>jsr305</artifactId>
+        <version>1.3.9</version>
+      </dependency>
```
xitep: don't we also have to add this dependency to every module utilizing `@Nullable` and friends?

vanekjar: You're right. I will fix this.

xitep: just came to my mind: alternatively, we might place this dependency directly under `<project> <build> <dependencies> ...` this will then make every child project dependent on it. this would save some typing, but is generally considered bad practice.

xitep:
```
[INFO] --- findbugs-maven-plugin:3.0.4:check (default) @ euphoria-fluent ---
[INFO] BugInstance size is 0
[INFO] Error size is 0
[INFO] No errors/warnings found
```

i like this! :) many thanks so far.

PR #49, opened 2017-03-11
Allow making fast builds

### LINKED ISSUE #48: make builds faster
Would it be possible to speed up `mvn install` for local development? It'd be great if I was able to turn off javadoc, findbugs, and possibly other build steps when developing locally. Maybe we can provide different  profiles such that we can opt out of certain build steps using `mvn install -P-xyz` when building interactively on the command line.

xitep:
* Moves most maven plugins under the profile "regular-build"
* _"regular-build"_ is active by default
* _"regular-build"_ can be deactivated on the cmdline by `mvn install -P-release-build`
* Resolves #48

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -151,10 +151,6 @@
             <source>1.8</source>
             <target>1.8</target>
             <encoding>UTF-8</encoding>
-            <compilerArgs>
-              <arg>-Xlint:unchecked</arg>
-              <arg>-Werror</arg>
-            </compilerArgs>
           </configuration>
         </plugin>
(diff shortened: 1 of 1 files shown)


vanekjar:
Cool! Works for me :+1:

je-ik:
Would it be possible to incorporate this change into the `Building` section of `README.md`? I know that's annoying, but I think that would be useful for many users. :)

xitep:
yep, good idea. thanks for the reminder. will add a note about it in a moment.

Issue #33, opened 2017-02-24
euphoria-flink: upgrade to flink 1.2.x


xitep:

xitep:
Already part of #27

Issue #27, opened 2017-02-17
Flink executor: Make use of ProcessFunction


vanekjar:
Recently released Flink 1.2.0 introduced a low-level API for streaming that gives access to all internal parts of the engine (data, state, timers).

> The ProcessFunction can be thought of as a FlatMapFunction with access to keyed state and timers. It handles events be being invoked for each event received in the input stream(s).

More details described at https://ci.apache.org/projects/flink/flink-docs-release-1.2/dev/stream/process_function.html

It seems it's a good opportunity to optimize current state of `windowing` in Flink stream executor in a way that all wrapping class from package `cz.seznam.euphoria.flink.streaming.windowing` can be discarded and replaced by custom `WindowOperator` implementation relying on Euphoria-core `windowing and triggers` only.

Issue #47, opened 2017-03-10
euphoria-flink: streaming word-count with time-sliding very slow


xitep:
Executing a word-count like program - this is the simplest scenario to trace down the problem - against an Euphoria's Flink Streaming executor using `TimeSliding` windowing, reveals a huge shuffle overhead compared to an implementation of the same program using Flink's native Java API.

### Experiment characteristics

* Apart from the key, the actual "values" being shuffled (as intended by the programmer) are actually small (in both cases):
  * On native flink: it's a word (string)
  * On euphoria: it's just a long (the value to be reduced)
* TimeSliding; the more windows the bigger Euphoria's overhead

### Problem

Euphoria tries to implement a strategy such that the size of the actual values being shuffled gets reduced. This is, instead of shuffling RBK's (and RSBK's) original input elements, before the shuffle phase it extracts the "values" (to be later reduced) and sends those for reduction (instead of the original input elements). The reasoning behind this choice is to send the typically smaller values instead of the bigger input elements. However, due to windowing being applied to the operator's original input elements, it also needs to extract the original input elements' windows before sending them together with the value across the wire. The more windows there are for a single value, the bigger the overhead of data being shuffled.

The more data is being shuffled, the quicker network buffers get filled, which can lead to back pressure, thereby making the consumer side go slower. (To mitigate against back pressure, we can resize the network buffers at the cost of memory consumption.)

Our implementation of the idea to shuffle only values instead of the original input elements suffers under the following items:
* If the "value" being shuffled isn't smaller than or equal size as the original input element, we're immediately introducing unnecessary overhead.
* If the "value" being shuffled _plus_ the necessary window information for the element isn't smaller than or equals size as the original input element, we're introducing unnecessary overhead. (note: this problem is not relevant when considering a partitioning schema by "window and key"; this is a potential but not-yet-implemented/explored reduce-by-key strategy.)
* Users of the Euphoria API have no chance to express a preference for a particular strategy in regards to the shuffle.

### What's to do next

- [x] Attempt to reduce the serialized size of window types natively provided by Euphoria
  * @vanekjar had a very nice idea about optimizing the `TimeSliding` windowing use-case: instead of creating many physical objects for each `TimeInterval` of the time slides, i.e. a `HashSet<TimeInterval>`, create one object which is able to lazily expand itself into these many window objects only when they are iterated - which for the RBK happens after the shuffle. the serialized form of such a "set of windows" can be of constant size independent of the number of windows presented
[...]

Issue #38, opened 2017-02-24
euphoria-core: unbounded source without explicit windowing defined


xitep:
Executing a _stateful_ operator which itself does not define a windowing strategy and which consumes - directly or indirectly - a _non-windowed, unbounded_ data set is basically undefined. This is, such an operator, e.g. `ReduceByKey`, consumes an infinite stream of data never reaching a point at which results can be emitted. Note:

* stateful: consuming non-windowed, unbounded data sources through _stateless_ operators, e.g. `FlatMap`, `Union`, `Repartition` _is_ defined. There's no problem for them. These emit results immediately upon receiving input.
* non-windowed and unbounded: this is, _all along the way from the unbounded data source down to the operator in question,_ there is no explicit windowing part of the game. since euphoria has the notion of "attached windowing", one windowing on the way to the operator in question, makes the processing defined again - even if that operator itself has no explicit windowing defined.

It'd be good to immediately fail the attempt to translate a flow with the described situation. The situation is mostly unintentional, but the mistake is hard to spot at runtime - basically leaving the programmer wondering why no output is produced.

One might argue, though, that the situation is practically valid (e.g. it works nicely on flink actually): the semantic of such a situation might well be that the results are produced when the unbounded data source is closed/cancelled. however, such computed results are non-deterministic and unsound with the theory.

je-ik:
I'm not sure if this is directly related to this - but there is at least one valid use case of stateful non-windowed operation - that is a non-windowed stream join. The example would look like this:
```java
  Dataset<?> stream = ...;
  Dataset<?> table = ...;
  Join.of(stream, table)
    .by(...)
    .using(...);    
```

This *can* work without windowing assuming that there is a possibility to convert the `table` into random access storage. This can be done in many ways, depending on the physical storage of the dataset (e.g. if it is a Kafka stream, then in can be cached locally in the way `KTable` works in `KStreams`). There is no windowing needed, because the stateful operation is effectively turned into a stateless, one-element-at-a-time operation.

xitep:
true enough; looks like the special case mentioned in #41

xitep:
[...]

PR #50, opened 2017-03-19
#27 [euphoria-flink] Rewrite windowing to native implementation of StreamOperator

vanekjar:
Finally, my PR is ready.

The code for handling windows is still over-complicated, but I deleted more code than actually added which is always a good sign.

Instead of using `ProcessFunction` I directly implemented custom `StreamOperator`. It turned out during implementation that `ProcessFunction` API is not so powerful for our use case - for example it's not possible to access the current `key` in callbacks.

There are a few pitfalls in the current PR that I am not very satisfied with:

 * Rewriting windowing haven't brought any performance benefit. Runtime is more or less the same as with previous implementation. On the other hand it now opens new opportunities to optimize - for example removing duplicate `timestamp` from `WindowedElement` flowing through the pipeline and use the built-in Flink timestamp instead. This will come in another PR soon.

 * I had to remove the functionality of flushing remaining windows in case of `bounded stream` end. So far all remaining windows were flushed to output in case of EOS. It proved that tracking of all existing windows had a huge performance drawback considering the registered window set must be persisted for fault-tolerance. I left that functionality only for unit testing (works for `TestFlinkExecutor` in local mode). Now it opens a discussion if we need this functionality of `bounded stream` in production.  

Thanks for the review.

* Resolves: #33 
* Closes: #27


euphoria-operator-testkit/src/main/java/cz/seznam/euphoria/operator/test/ReduceByKeyTest.java
```diff
@@ -747,11 +747,11 @@ public void validate(Partitions<Integer> partitions) {
 
         ArrayList<Long> triggerTimes = new ArrayList<>(TETETS_SEEN_TIMES_TRIGGER);
         triggerTimes.sort(Comparator.naturalOrder());
-        assertEquals(asList(15_000L, 19_999L, 25_000L, 29_999L), triggerTimes);
+        assertEquals(asList(19_999L, 19_999L, 29_999L, 29_999L), triggerTimes);
 
         ArrayList<Long> assignerTimes = new ArrayList<>(TETETS_SEEN_TIMES_ASSIGNER);
         assignerTimes.sort(Comparator.naturalOrder());
-        assertEquals(asList(15_000L, 19_999L, 25_000L, 29_999L), assignerTimes);
+        assertEquals(asList(19_999L, 19_999L, 29_999L, 29_999L), assignerTimes);
```
xitep: not sure we need this test anymore. i think `testElementTimestamp` should suffice.

euphoria-flink/src/test/java/cz/seznam/euphoria/flink/streaming/windowing/KeyedMultiWindowedElementTest.java
```diff
@@ -0,0 +1,37 @@
...
+public class KeyedMultiWindowedElementTest {
+
+  @Test
+  public void testSerializer() {
+    TypeSerializer<KeyedMultiWindowedElement> serializer =
+            TypeInformation.of(KeyedMultiWindowedElement.class)
+                    .createSerializer(new ExecutionConfig());
+
+    // must be POJO serializer for performance reasons
+    assertTrue(serializer instanceof PojoSerializer);
```
xitep: thanks! :+1:

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/ReduceByKeyTranslator.java
```diff
@@ -1,270 +0,0 @@
...
-import org.apache.flink.streaming.api.functions.windowing.WindowFunction;
-import org.apache.flink.streaming.api.windowing.windows.Window;
-import org.apache.flink.util.Collector;
-
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.Set;
-
-class ReduceByKeyTranslator implements StreamingOperatorTranslator<ReduceByKey> {
```
xitep: no specialization for the `ReduceByKey` operator in the streaming executor anymore?

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/StreamingWindowedElement.java
```diff
@@ -1,33 +0,0 @@
...
- * of this type.
- */
-public class StreamingWindowedElement<W extends Window, T>
-    extends WindowedElement<W, T>
-    implements ElementProvider<T>
-{
-  public StreamingWindowedElement(W window, long timestamp, T element) {
-    super(window, timestamp, element);
-  }
-}
```
xitep: many thanks for the clean-up!

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/ReduceByKeyTranslator.java
vanekjar: It's easier to maintain just one implementation. Flink is internally doing the same job with incremental reducer as we do now in `WindowOperator`. 
Anyway the performance of benchmark using `ReduceByKey` is the same with the new implementation.

xitep: i agree it's easier to maintain only one implementation. no doubt about that. what i worry about a bit is the difference between serializing the state (into the backend storage) of a combining RBK (a single value) vs. serializing the same value in a list (of size one.) my guess is, we cannot tell this difference now, since other factors are likely to hide this overhead. let's stick with one implementation and optimize later - if necessary.

vanekjar: I agree with you concern about difference between `ValueStorage` and `ListStorage`. That can really make a difference. 
There is a chance to optimize the internals of `ReduceByKey.ReduceState` in a way that it will use `ValueStorage` in case it's `combinable`. Also the storage is accessed (serialized) multiple times now when adding to state, this can be also avoided.

xitep: if necessary, i think we can even optimize this later directly in the api layer in `#getBasicOps` of the `RBK` operator. this would allow us to specialize the case without having to maintain separate impls in the executors.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/ReduceStateByKeyTranslator.java
```diff
@@ -44,114 +47,124 @@
...
 
+    // assign windows
+    DataStream<KeyedMultiWindowedElement> windowed = input.map(new WindowAssigner(
+            windowing, keyExtractor, valueExtractor, eventTimeAssigner))
+            .setParallelism(operator.getParallelism());
+
+    DataStream<WindowedElement<?, Pair>> reduced = (DataStream) windowed.keyBy(new KeyExtractor())
+            .transform(operator.getName(), TypeInformation.of(WindowedElement.class), new WindowOperator<>(
+                    windowing, stateFactory, stateCombiner, context.isLocalMode()))
+            .setParallelism(operator.getParallelism());
```
xitep: i would like to see support for running this "as is", as well as with the value extraction and window assignment functionality executing only after the shuffle (maybe some global parameter to the translation layer). at least for our benchmarking this will be necessary. what do you think it would take to support both?

vanekjar: Definitely agree it would be useful. But I am not sure if this is not the part of #47

xitep: oh yeah, right. let's introduce that with the mentioned ticket later.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/windowing/WindowOperator.java
```diff
@@ -0,0 +1,465 @@
...
+                  // FIXME This implementation relies on fact that
+                  // stateCombiner function is actually "fold left".
+                  // That means the result state will end up in the
+                  // first state given to the combiner function.
+
+                  // merge all mergedStateWindows into stateResultWindow
+                  List<State> states = new ArrayList<>();
+                  states.add(getWindowState(stateResultWindow));
+                  mergedStateWindows.forEach(sw -> states.add(getWindowState(sw)));
+                  stateCombiner.apply(states);
```
xitep: in regards to the above FIXME, i wanted to suggest to change the type of the state-combiner. however, now i see it's a `CombinableReduceFunction` which does have a return value that is supposed to replace the merged states. in the `inmem` as well as in the spark executor (e.g. [GroupedReducer](https://github.com/seznam/euphoria/blob/master/euphoria-core/src/main/java/cz/seznam/euphoria/core/executor/greduce/GroupReducer.java#L182)) we are doing so. if i'm not mistaken applying the same technique here should resolve the above FIXME.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/windowing/MergingWindowSet.java
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * Utility class to keep track of currently existing {@link Window} instances
+ * when using {@link MergingWindowing}.
+ *
+ * @param <W> type of the window
+ */
+class MergingWindowSet<W extends Window> {
+
+  private final MergingWindowing windowing;
```
xitep: this is quite cool :sunglasses:! do you think we can re-use it to replace parts of `GroupedReducer` in `euphoria-core`?

xitep: ah, github is playing tricks on me ;) ... that comment was meant to address the class as a whole.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/windowing/AttachedWindowing.java
```diff
@@ -0,0 +1,68 @@
...
+    return new AttachedWindowTrigger<>();
+  }
+
+  private static class AttachedWindowTrigger<WID extends Window> implements Trigger<WID> {
+
+    @Override
+    public TriggerResult onElement(long time, WID window, TriggerContext ctx) {
+      // FIXME batch window shouldn't be used in stream flow in the future
+      // issue #38 on GitHub
+      if (window instanceof Batch.BatchWindow) return TriggerResult.NOOP;
```
xitep: what about throwing an exception here?

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/ExecutionEnvironment.java
```diff
@@ -22,6 +22,7 @@
 import cz.seznam.euphoria.core.client.dataset.windowing.WindowedElement;
 import cz.seznam.euphoria.core.client.flow.Flow;
 import cz.seznam.euphoria.core.client.util.Pair;
+import cz.seznam.euphoria.flink.streaming.windowing.KeyedMultiWindowedElement;
```
xitep: looks like an unused import

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/ReduceStateByKeyTranslator.java
```diff
@@ -44,114 +47,124 @@
...
+      input = input.assignTimestampsAndWatermarks(
+              new EventTimeAssigner(context.getAllowedLateness(), eventTimeAssigner));
     }
 
+    // assign windows
+    DataStream<KeyedMultiWindowedElement> windowed = input.map(new WindowAssigner(
+            windowing, keyExtractor, valueExtractor, eventTimeAssigner))
+            .setParallelism(operator.getParallelism());
+
+    DataStream<WindowedElement<?, Pair>> reduced = (DataStream) windowed.keyBy(new KeyExtractor())
```
xitep: i think we're better off with `.keyBy("key")`; flink will then derive the type information from the input data stream automatically; it'll still be "object" at this moment since we don't supply enough type information through the `WindowAssigner`, but that's about to come in some future.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/windowing/WindowOperator.java
vanekjar: I am not sure about that. The returned state from `CombinableReduceFunction` doesn't matter at all. Since our state is basically "stateless", it depends if the resulting state stored the merged value to the appropriate persistent storage using correct namespace.
In this case we need the state combiner to store the result in `stateResultWindow` namespace. And I can't see any method how to ensure that.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/windowing/AttachedWindowing.java
vanekjar: Unfortunately there still exists unit test with unbounded source and no windowing. I think it needs to be resolved with the issue #38

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/windowing/WindowOperator.java
xitep: ah, thanks for the explanation! now i see it. we'll follow up on this later - combinable-reduce-function then doesn't seem right to me at this place. may i ask you to set up a ticket for that FIXME?

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/windowing/AttachedWindowing.java
xitep: ok, fine with me.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/windowing/WindowOperator.java
```diff
@@ -0,0 +1,465 @@
...
+    WID stateWindow = window;
+    State windowState;
+
+    if (windowing instanceof MergingWindowing) {
+      Objects.requireNonNull(mergingWindowSet);
+      stateWindow = mergingWindowSet.getStateWindow(window);
+      windowState = getWindowState(stateWindow);
+    } else {
+      windowState = getWindowState(window);
+    }
```
xitep: potential micro-optimization possibility here: don't lookup out the window state from the state table when not necessary (e.g. `tr == NOOP`)

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/windowing/WindowOperator.java
vanekjar: Issue created #51

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/ReduceStateByKeyTranslator.java
vanekjar: Great idea, but unfortunately doesn't work. Or at least I don't know how make it work. Object is not a key type.
```
org.apache.flink.api.common.InvalidProgramException: This type (GenericType<java.lang.Object>) cannot be used as key.
```

xitep: hm ... we'll need more type information :/ anyway, thanks for having a try!

xitep:
thank you so much for cleaning up old stuff! feel free to merge into master. it looks very good to me.

the fact that the performance didn't get better means we didn't get worse! :) (our benchmark is just a single scenario and it happens to hit a particular bottleneck which is present in both versions.) it be interesting to compare them without the bottleneck eliminated.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/windowing/MergingWindowSet.java
vanekjar: I don't think reusing this class would bring any benefit. It's too complicated because the window set must be persisted after each step in streaming. Most of the code is performance optimization to avoid costly persistent state allocation. This is not the case in `GroupReducer` where everything is processed in a plain `HashMap` in memory.

PR #52, opened 2017-03-21
#! [euphoria-flink] Avoid extra shuffle when windowing on streaming

xitep:
Moves the map function into the operator chain being executed before `.keyBy` avoiding an extra re-balancing operation.

### DIFF
--- a/euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/ReduceStateByKeyTranslator.java
+++ b/euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/ReduceStateByKeyTranslator.java
@@ -73,7 +73,9 @@ class ReduceStateByKeyTranslator implements StreamingOperatorTranslator<ReduceSt
     // assign windows
     DataStream<KeyedMultiWindowedElement> windowed = input.map(new WindowAssigner(
             windowing, keyExtractor, valueExtractor, eventTimeAssigner))
-            .setParallelism(operator.getParallelism());
+            // ~ execute in the same chain of the input's processing
+            // so far, thereby, avoiding an unnecessary shuffle
+            .setParallelism(input.getParallelism());
 
     DataStream<WindowedElement<?, Pair>> reduced = (DataStream) windowed.keyBy(new KeyExtractor())
             .transform(operator.getName(), TypeInformation.of(WindowedElement.class), new WindowOperator<>(


xitep:
I'm sorry, I didn't spot this yesterday as part of the review for #50. Here's the execution graph with the extra re-balancing:

testapps-graph

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/ReduceStateByKeyTranslator.java
```diff
@@ -73,7 +73,9 @@
     // assign windows
     DataStream<KeyedMultiWindowedElement> windowed = input.map(new WindowAssigner(
             windowing, keyExtractor, valueExtractor, eventTimeAssigner))
-            .setParallelism(operator.getParallelism());
+            // ~ execute in the same chain of the input's processing
+            // so far, thereby, avoiding an unnecessary shuffle
+            .setParallelism(input.getParallelism());
```
je-ik: What is the difference of `input.getParallelism()` and `operator.getParallelism()`? Shouldn't this be the same provided that to operator didn't define its own parallelism?

xitep: * `input.getParallelism()` returns the parallelism the corresponding flink's dataset
* `operator.getParallism()` returns the _target_ parallelism intended for the operator (in this case, rsbk)

je-ik: Yes, okay, but if the user code does not define specific parallelism (for RSBK), shouldn't this be the same?

xitep: ah, yes. that's correct! in that case `input.parallelism` will equal `operator.parallelism`, that's right.

je-ik: Okay, I didn't look into details of this PR, but will the RSBK respect the user-supplied parallelism (if any) after this change?

xitep: yes, it will. the attached execution plan shows a program where the RSBK is instructed for a 120 partitions output.

vanekjar:
Good catch! Thanks :+1:

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/ReduceStateByKeyTranslator.java
vanekjar: Yes, user-supplied parallelism is correctly applied to the operation after shuffle. It's basically `map -> keyBy -> setParallelism -> reduceByKey`.

PR #57, opened 2017-03-23
Drop WindowedElementImpl from euphoria-core

xitep:


xitep:
These are changes we discussed offline. Feel free to merge #56 after merging this one. And many thanks for the `WindowedElement` interface!

vanekjar:
Thanks a lot :+1:

Issue #14, opened 2017-02-01
Performance tuning of Flink stream executor


vanekjar:
This ticket is considered to be a child of the story #11, where more detailed description of the problem can be found.

je-ik:
Obsolete.

PR #59, opened 2017-03-24
#14 [euphoria-flink] Don't send timestamp along with each element.

vanekjar:
In Flink streaming executor a timestamp is not sent with each element anymore. Instead it relies on internal Flink timestamps.

Seems it introduces noticeable performance benefits.

### DIFF
--- a/euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/InputTranslator.java
+++ b/euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/InputTranslator.java
@@ -20,7 +20,6 @@ import cz.seznam.euphoria.core.executor.FlowUnfolder;
 import cz.seznam.euphoria.flink.FlinkOperator;
 import cz.seznam.euphoria.flink.streaming.io.DataSourceWrapper;
 import org.apache.flink.streaming.api.datastream.DataStream;
-import org.apache.flink.streaming.api.functions.IngestionTimeExtractor;
 
 class InputTranslator implements StreamingOperatorTranslator<FlowUnfolder.InputOperator> {
 
@@ -33,7 +32,6 @@ class InputTranslator implements StreamingOperatorTranslator<FlowUnfolder.InputO
 
     return context.getExecutionEnvironment()
             .addSource(new DataSourceWrapper<>(ds))
-            .setParallelism(operator.getParallelism())
-            .assignTimestampsAndWatermarks(new IngestionTimeExtractor<>());
+            .setParallelism(operator.getParallelism());
   }
 }
--- a/euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/ReduceStateByKeyTranslator.java
+++ b/euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/ReduceStateByKeyTranslator.java
@@ -37,6 +37,7 @@ import org.apache.flink.api.java.typeutils.ResultTypeQueryable;
 import org.apache.flink.streaming.api.datastream.DataStream;
 import org.apache.flink.streaming.api.functions.timestamps.BoundedOutOfOrdernessTimestampExtractor;
 import org.apache.flink.streaming.api.operators.AbstractStreamOperator;
+import org.apache.flink.streaming.api.operators.ChainingStrategy;
 import org.apache.flink.streaming.api.operators.OneInputStreamOperator;
 import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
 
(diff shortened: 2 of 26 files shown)


euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/windowing/WindowAssigner.java
```diff
@@ -0,0 +1,105 @@
...
+  @SuppressWarnings("unchecked")
+  public KeyedMultiWindowedElement<W, KEY, VALUE> apply(StreamRecord<StreamingElement<?, I>> record) {
+    StreamingElement el = record.getValue();
+
+    if (reuse == null) {
+      reuse = new TimestampedElement();
+    }
+    reuse.setTimestamp(record.getTimestamp());
+    reuse.setElement(el);
+    Set windows = windowing.assignWindowsToElement(reuse);
```
xitep: i wonder if we shouldn't call `reuse.setElement(null)` after the call to `assignWindowsToElement` to avoid referencing the last processed streaming element. but it probably makes no difference.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/windowing/WindowAssigner.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Override
+    public T getElement() {
+      return element.getElement();
+    }
+
+    public void setElement(StreamingElement<W, T> element) {
+      this.element = element;
+    }
```
xitep: this doesn't look good to me :/ at least it's surprising from a code perspective. the getter returns some type `T` while the setter takes a `StreamElement<W, T>` ... can we maybe call the setter explicitly `setStreamingElement`?

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/windowing/WindowAssigner.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  private static class TimestampedElement<W extends Window, T>
+          implements WindowedElement<W, T> {
+
+    private long timestamp;
+    private StreamingElement<W, T> element;
+
+    public TimestampedElement() {
+    }
```
xitep: i guess we can drop this and thus shorten the code a little bit.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/ReduceStateByKeyTranslator.java
```diff
@@ -137,47 +141,28 @@ public long extractTimestamp(FlinkElement element) {
...
-                          UnaryFunction valueExtractor,
-                          UnaryFunction eventTimeAssigner) {
-      this.windowing = windowing;
-      this.keyExtractor = keyExtractor;
-      this.valueExtractor = valueExtractor;
-      this.eventTimeAssigner = eventTimeAssigner;
+    private final WindowAssigner windowAssigner;
+
+    private WindowAssignerOperator(WindowAssigner windowAssigner) {
+      this.windowAssigner = windowAssigner;
```
xitep: I've seen Flink's `StreamMap` (which was previously used for the work done in this new `WindowAssiginerOperator`) uses `chainingStrategy = ChainingStrategy.ALWAYS;` while the default value is `.HEAD`) Can you try that out and see if it has any effect on the executed DAG?

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/InputTranslator.java
```diff
@@ -33,7 +32,6 @@
 
     return context.getExecutionEnvironment()
             .addSource(new DataSourceWrapper<>(ds))
-            .setParallelism(operator.getParallelism())
-            .assignTimestampsAndWatermarks(new IngestionTimeExtractor<>());
+            .setParallelism(operator.getParallelism());
```
vanekjar: I am afraid it doesn't. We're still running using `ingestion` time characteristics, not `processing`. So the periodic watermarks will be send.
Also I think testing of `ingestion` or `processing` may be quite tricky because of flipping tests.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/windowing/WindowAssigner.java
vanekjar: Good observation! This appeared as a result of some name refactoring. Thanks :+1:

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/ReduceStateByKeyTranslator.java
vanekjar: Cool! This indeed helped to remove one stage/shuffle from the DAG resulting in even better performance. Thanks :+1:

xitep:
Thanks!

PR #62, opened 2017-03-28
#! Prefer ValueStorage instead of ListStorage for combining RSBK operations

### LINKED ISSUE #60: euphoria-flink: optimize rocks db access for combining rbk
A _combining_ RBK operator currently does `state.get(); state.clear(); state.add();` to store its running state value. The call to `state.clear()` is unnecessary. Ideally, we'd replace `state.clear(); state.add()` with `state.set()`, which should be more efficient.

* This change can and should be achieved in the default implementation of RBK; see `RBK#getBasicOps`. `accumulate(); state.set(..);` much better expresses the intent and allows for an optimal implementation.
* It is applicable only for the _combining_ scenario.
* We shall measure and document the impact of the change.

xitep:
* Resolves #60

### DIFF
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/ReduceByKey.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/ReduceByKey.java
@@ -313,7 +313,7 @@ public class ReduceByKey<
 
       @Override
       public State<E, E> apply(Context<E> ctx, StorageProvider storageProvider) {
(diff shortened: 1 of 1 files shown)


xitep:
This seems to make a big difference (in our benchmark, since we leverage the new impl in two stages.) I'm still investigating around this change, through I would greatly appreciate an early, critical review.

vanekjar:
Looks cool to me! Definitely much better performance wise :+1:

PR #66, opened 2017-03-29
#47 TimeSliding window assigner generates custom SlidingWindowSet

vanekjar:
This PR reduces shuffle data size when using time sliding windowing by approximately 90 % according to benchmark measurement.

Instead of shuffling a collection of `TimeInterval` instances with each element, custom class `SlidingWindowSet` consisting of 3 `long` numbers  (start, end, slide interval) is used.


xitep:
muchos gracias!

PR #71, opened 2017-04-04
#51 [euphoria-core] Introduce StateMerger

### LINKED ISSUE #51: ReduceStateByKey state combiner function shouldn't use CombinableReduceFunction
Currently the state combiner is type of `CombinableReduceFunction`

```java
public State apply(Iterable<State> reducedStates)
```

It turned out runtime needs to choose which of given states will be the final result merged into. That means the function can't be commutative.

Proposed solution is to change the signature of the state combiner to something like
```java
public void merge(State targetState, Iterable<State> mergedStates)
```

This task involves change of client API as well as modification of all executors relying on this API.

xitep:
* Resolves #51


euphoria-core/src/main/java/cz/seznam/euphoria/core/executor/greduce/GroupReducer.java
```diff
@@ -183,6 +185,18 @@ private void onTimerCallback(long stamp, WID window) {
...
       }
 
+      // ~ as of now on, we can assume `sources` do _not_ contain `target`;
+      // this implies:
+      //  a) the target window's state will not be merged into itself
+      //  b) the target window's triggers will not be merged into itself
+      //  c) the target window's trigger #onClear won't be called
+      sources.remove(target);
+
+      // XXX only if sources non empty!
```
xitep: indeed, it's a TODO i forgot about. thanks for pointing it out! yes, skipping/continuing can/should be done. will fix this.

PR #74, opened 2017-04-05
#19 [euphoria-flink] Avoid extra shuffle (on the batch executor)

### LINKED ISSUE #19: euphoria-flink: reduce number of required shuffles
In particular, we'd like to resolve this FIXME in the batch and streaming `ReduceStateByKeyTranslator` in `euphoria-flink` to reduce the number of shuffle operations of the operator implemenation:

```
    // FIXME parallelism should be set to the same level as parent until we reach "shuffling"
```

xitep:
* Resolves #19


xitep:
Given the flow `Input.withPartitions(1) -> FlatMap -> ReduceByKey.withPartitions(2) -> ReduceByKey.withPartitions(1) -> ...` here's the execution graf on current master:

baseline

And here's the same with the patch applied:

patched

vanekjar:
Looks good to me! Thanks :+1:

PR #78, opened 2017-04-08
State consistency

### LINKED ISSUE #70: euphoria-flink: state storage consistency
As part of #67 we observed an uncertainty about where trigger states get stored for merging windows. After windows get merged, one of the already existing, merged windows becomes the carrier of the target window's state (to avoid shuffling state from one window into a new one.) However, it appears that the `onEventTime` method utilizes partially the target window's namespace as well as the state carrier window's namesspace to handle triggers and their states.

* Please clarify
* Add a test to validate that trigger state for merging windows is at exactly one place

xitep:
* Resolves #70


vanekjar:
I agree with the fix. Nice unit test covering this bug :+1: 
It is still questionable if the trigger state should be stored in different scope than the window state. That means trigger state is moved after each window merge to the target window. In contrast window state is stable because of optimization in `MergingWindowSet`. But I think we can live with that for now.

PR #81, opened 2017-04-10
#61 Allowed lateness unit test

### LINKED ISSUE #61: euphoria-flink: Allowed lateness
In streaming environment it happens elements arrive out-of-order. Our `FlinkExecutor` has a possibility to setup a time period called `allowed lateness` via `setAllowedLateness` method. Allowed lateness specifies by how much time elements can be late before they are dropped.

Currently `allowed lateness` is applied during event time assignment in `ReduceStateByKeyTranslator.EventTimeAssigner`. In case no event time extraction is used (i. e. flow is run using `ingestion` time characteristics) the `allowed lateness` is ignored. It would be better to move `allowed lateness` to `AbstractWindowOperator` and remove it from assigner.

vanekjar:
It seems we're handling late-coming records correctly. `Allowed lateness` is intended to be applied when extracting `event time` from elements. It is not necessary to handle out-of-order elements (late-comers) in case of `ingestion time` since this time characteristics is by definition increasing.

I just added unit test covering the case with out-of-order elements to make sure we're handling this situation correctly.

closes #61

### DIFF
--- a/euphoria-flink/src/test/java/cz/seznam/euphoria/flink/streaming/RBKTimeWindowTest.java
+++ b/euphoria-flink/src/test/java/cz/seznam/euphoria/flink/streaming/RBKTimeWindowTest.java
@@ -226,5 +226,50 @@ public class RBKTimeWindowTest {
         .collect(Collectors.toList());
   }
 
+  @Test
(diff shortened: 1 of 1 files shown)


xitep:
Merci!

PR #82, opened 2017-04-10
#79 [euphoria-core] Trigger#onMerge does not return TriggerResult

### LINKED ISSUE #79: Bugs in Merging Windowing process
When I was going through the code handling window merging I realized there must clearly a few bugs. Although it is not covered by unit test. I added more complex unit test in branch [79/MergingWindowingBug](https://github.com/seznam/euphoria/compare/79/MergingWindowingBug) that is currently failing.

The test uses merging windowing that is uncommon in two attributes:

 * Windows are always triggered only by `onMerge` trigger.
 * Newly created window is never result of the merge, it just triggers the merge on other windows.

Running this test will fail on both executors:

 * In Flink `AbstractWindowOperator` flushes the window `onMerge` even before the window state is merged to the final result. That leads to incorrect result. 
https://github.com/seznam/euphoria/blob/master/euphoria-flink/src/main/java/cz/seznam/euphoria/flink/streaming/windowing/AbstractWindowOperator.java#L173

 * In Spark `GroupReducer` ignores the trigger result of `onMerge` operation in case the newly created window is not part of the merge result.
https://github.com/seznam/euphoria/blob/master/euphoria-core/src/main/java/cz/seznam/euphoria/core/executor/greduce/GroupReducer.java#L218

Please have a look and let me know if my observation is right.

vanekjar:
closes #79

### DIFF
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/triggers/AfterFirstCompositeTrigger.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/triggers/AfterFirstCompositeTrigger.java
@@ -75,13 +75,9 @@ public class AfterFirstCompositeTrigger<W extends Window> implements Trigger<W> 
   }
 
   @Override
-  public TriggerResult onMerge(W window, TriggerContext.TriggerMergeContext ctx) {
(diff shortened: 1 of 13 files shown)


xitep:
Thank you!

PR #83, opened 2017-04-11
Hadoop string sink

xitep:


xitep:
@horkyada as we discussed offline, i generalized the sink, may i ask you to approve this pr?

PR #84, opened 2017-04-11
#! [euphoria-flink] Warn about failing execution scenario

xitep:


vanekjar:
Nice one! Thanks for reporting it to Flink JIRA :+1:

PR #96, opened 2017-04-18
Num output partitions for multi-input operators

### LINKED ISSUE #92: Output partitioning of Union operator is not well defined
Number of output partitions (as reflected by `Dataset#getNumPartitions()`) of `Union` operator should probably be one of the following:
 a) sum of partitions of input `Dataset`s
 b) maximum of number of input partitions over input Dataset

Currently, it seems that the partitioning is taken from the first input `Dataset`, which seems to be wrong. I'm not 100% sure about the consequences of changing the semantics, so I'm leaving this for a discussion under this issue.

xitep:
* Sets `#getNumPartitions() == sum(input.getNumPartitions()` for the two multi input operators we have, i.e. `Union` and `Join`
* An explicit test in the operator-testkit seems no necessary due to the change being guaranteed by `euphoria-core` already
* Resolves #92

### DIFF
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/dataset/Datasets.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/dataset/Datasets.java
@@ -53,6 +53,52 @@ public class Datasets {
   }
 
   /**
+   * Create an output dataset for the given operator, assuming it is consuming
+   * both given inputs.
+   *
+   * @param <IN> the type of the input elements
+   * @param <OUT> the type of the operators output elements
+   *
+   * @param flow the flow the operator and the inputs are part of
+   * @param one one of the operators inputs
+   * @param two the other input of the operator
+   * @param op the operator consuming the two given inputs
+   *
+   * @return a dataset representing the output of the specified operator
+   */
+  public static <IN, OUT> Dataset<OUT> createOutputForBinaryInput(
+      Flow flow, Dataset<IN> one, Dataset<IN> two, Operator<IN, OUT> op) {
+    if (one.isBounded() != two.isBounded()) {
+      throw new IllegalStateException("Cannot combine unbounded and bounded inputs!");
+    }
+    return new OutputDataset<OUT>(flow, op, one.isBounded()) {
+      @Override
+      public int getNumPartitions() {
+        if (op instanceof PartitioningAware) {
+          return ((PartitioningAware) op).getPartitioning().getNumPartitions();
+        } else {
+          return combinedNumPartitions(one, two);
+        }
+      }
+    };
+  }
+
+  /**
+   * Determines the number of partitions when combining the two given inputs.
+   *
+   * @param x one of the input datasets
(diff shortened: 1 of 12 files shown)


je-ik:
Many thanks for solving this issue. I wonder if there is really no need to incorporate this into operator test-kit, do you see any chance that any executor might choose to ignore the `Dataset#getNumPartitions()` setting, and instead implement some other partitioning strategy? I think that we should make sure that no executor does this (and these modifications of `euphoria-core` does not really prevent this, or does?).
Thanks for comments, I will resolve this PR right after that.

xitep:
indeed! an additional test unveiled the unexpected behaviour on flink :) thanks for pushing! however, i'm not sure our decision regarding the "sum" is correct. at least on flink it introduces one additional shuffle:

```java
        Dataset<Integer> first = flow.createInput(ListDataSource.of(bounded,
            Arrays.asList(1)).withReadDelay(Duration.ofMillis(1000)));

        Dataset<Integer> second = flow.createInput(ListDataSource.of(bounded,
            Arrays.asList(2, 3),
            Arrays.asList(4, 5, 6)).withReadDelay(Duration.ofMillis(10)));

        Dataset<Integer> combined = Union.of(first, second).output();
        return MapElements.of(combined).using(e -> e).output();
```

before (i.e. `union.output.numPartitions = max(inputs.numPartitions)`):

_before

after (i.e. `union.output.numPartitions = sum(inputs.numPartitions)`):

_after

to avoid the shuffles/rebalancing on flink, we'd probably need to duplicate and "attach" the map outselves onto the source while translating the flow.

je-ik:
Oops, I'm not sure if I understand correctly, but does this mean that we cannot enforce this behavior in flink without extra repartitioning? If it is really impossible, then I'd say we would have to redefine the partitioning again - and let it be executor-defined (somehow, don't have an exact idea on how to do it).

xitep:
i'm sorry, but i fail to avoid the additional rebalancing on flink. if we'd like to keep the "sum" contract we might want to try out the idea of "duplicting and attaching the consumer onto the sources" (but this won't be easy.)

je-ik:
That sounds like we cannot guarantee number of output partitions in the `euphoria-core` and it would be best to somehow postpone this decision to the executor. But that also implies that the `Dataset#getNumPartitions()` is pretty much undefined (if the number of partitions is not defined by the user code explicitly). Shouldn't we just return `-1` for these cases?

xitep:
yes, i believe it would good to give executors a chance to optimize the execution as best as they can, i.e. to leave the aspect undefined in `euphoria-core`, unless the user specifies a desired partitioning. `-1` sounds good to me, but then, i wonder whether we need `Dataset#getNumPartitions` at all.

je-ik:
Maybe you're right. Can we drop it then, so that it is not confusing for users? Should I close this and make another issue?

xitep:
i'll set up a different PR, but we can keep the original ticket. many thanks for brain storming so far :+1:

PR #99, opened 2017-04-19
Adam/40/sort

horkyada:


xitep:
@horkyada may i ask you to rebase against current master?

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Join.java
```diff
@@ -442,7 +442,7 @@ public boolean isOuter() {
         OUT, JoinState, W> reduce;
 
     name = getName() + "::ReduceStateByKey";
-    reduce = new ReduceStateByKey<>(
+    reduce = new ReduceStateByKey(
```
xitep: what has changed that it suddenly started being a problem?

horkyada: It's just an eclipse problem. I included these chnages (~5-10) inside this PR, should I open special one for this?

xitep: no problem. just keep it here. that's fine with me.

is there any chance an upgrade to eclipse would solve the issue (given that `javac` has no problem compiling it i'd expect there's a particular issue with the eclipse compiler?) it seems worth being traced back to the root-cause and documented (maybe in the wiki)

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/SortTranslator.java
```diff
@@ -0,0 +1,156 @@
...
+      if (eventTimeAssigner != null) {
+        wel.setTimestamp(eventTimeAssigner.extractTimestamp(wel.getElement()));
+      }
+
+      Iterable<Window> windows = windowing.assignWindowsToElement(wel);
+      List<Tuple2<Tuple3<Integer, Window, Comparable>, Object>> out = new ArrayList<>();
+      for (Window wid : windows) {
+        Object el = wel.getElement();
+        out.add(new Tuple2<>(
+            new Tuple3(keyExtractor.apply(el), wid, sortByFn.apply(el)), wel));
```
vanekjar: I am afraid it's wrong here. Output element (`wel`) carries the original window, not the newly assigned. It seems this situation can be solved the same way as in `ReduceByKeyTranslator` by extracting `TimestampedElement` and reconstructing `SparkElement` in the final map operation.
Also it shows windowing is not properly covered in operator-test-kit.

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Sort.java
```diff
@@ -0,0 +1,303 @@
...
+  private static class PartitionKeyExtractor<IN, S extends Comparable<S>> implements UnaryFunction<IN, Integer> {
+
+    private final UnaryFunction<IN, S> sortByFn;
+    private final Partitioner<S> partitioner;
+    private final int numPartitions;
+    
+    public PartitionKeyExtractor(UnaryFunction<IN, S> sortByFn, Partitioning<S> partitioning) {
+      this.sortByFn = sortByFn;
+      this.partitioner = partitioning.getPartitioner();
+      this.numPartitions = partitioning.getNumPartitions();
```
xitep: yes, this will not work. at the moment i'm a bit clueless how to go about it.

@horkyada may i ask you to add a corresponding comment to #92 so that we consider and fix it?

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Sort.java
```diff
@@ -0,0 +1,303 @@
...
+            String name,
+            Dataset<IN> input,
+            UnaryFunction<IN, S> sortByFn,
+            Partitioning<S> partitioning,
+            @Nullable Windowing<IN, W> windowing,
+            @Nullable ExtractEventTime<IN> eventTimeAssigner) {
+    super(name, flow, input, 
+        new PartitionKeyExtractor<>(sortByFn, partitioning), 
+        windowing, eventTimeAssigner, 
+        new HashPartitioning<>(partitioning.getNumPartitions()));
```
xitep: yes, please document this. i spend quiet some time on this!

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Sort.java
```diff
@@ -0,0 +1,305 @@
...
+                eventTimeAssigner,
+                (StateFactory<IN, IN, Sorted<IN>>)
+                    (ctx, provider) -> new Sorted<>(ctx, provider, new SortByComparator<>(sortByFn)),
+                stateCombiner,
+                partitioning);
+
+    MapElements<Pair<Integer, IN>, IN>
+        format =
+        new MapElements<>(getName() + "::MapElements", flow, reduce.output(),
+            e -> e.getSecond());
```
xitep: prefer `Pair::getSecond` here please

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Sort.java
xitep: hm, looking more deeply into this, it seems to me that what we did as part of #96 will simply be necessary - i'm considering re-opening that pr again; ping @je-ik

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/batch/SortTranslator.java
```diff
@@ -0,0 +1,98 @@
...
+            c.collect(new BatchElement<>(wid, wel.getTimestamp(), el));
+          }
+        })
+        .returns(BatchElement.class)
+        .name(operator.getName() + "::map-input")
+            .setParallelism(inputParallelism);
+    
+    // ~ repartition and sort partitions
+    DataSet<BatchElement> sorted = wAssigned
+        .partitionCustom(new PartitionerWrapper<>(
```
xitep: if i'm not mistaken, we do `elem.getPartition() & Integer.MAX_VALUE) % numPartitions` twice; once as part of `PartitionerWrapper` and once as part of the hard-coded `Sort.PartitionKeyExtractor`, correct? if so, i would rather vote for changing the wrapper here to merely validate that the partition-id supplied to by the wrapped partitioner is in the correct range.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/batch/SortTranslator.java
```diff
@@ -0,0 +1,98 @@
...
+            origOperator.getPartitioning().getPartitioner()),
+            Utils.wrapQueryable(we -> udfKey.apply(we.getElement()), Integer.class))
+        .setParallelism(operator.getParallelism())
+        .sortPartition(Utils.wrapQueryable(
+            (BatchElement we) -> Tuple2.of(we.getWindow(), udfSort.apply(we.getElement())),
```
xitep: * i think this will again be incredibly slow. `Window` is an opaque type and flink will have to deserialize on every comparison i guess. or am i missing something?
* further, even in the case of knowing the actual window type and being able to register it with flink, we'd also need to make our `Window` impls POJOs or resort to registering special serializers and comparators for them.

euphoria-inmem/src/test/java/cz/seznam/euphoria/inmem/InMemExecutorTest.java
```diff
@@ -337,6 +337,11 @@ public boolean equals(Object o) {
     public int hashCode() {
       return size;
     }
+
+    @Override
+    public int compareTo(SizedCountWindow o) {
+      return 0;
+    }
```
xitep: i think this is a nice example showing how _not_ to implement `compareTo` for a window; please make it consistent with the `equals` impl at least for the sake of demo purposes.

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/SortTranslator.java
xitep: wow! good catch! 👍

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/dataset/partitioning/RangePartitioning.java
```diff
@@ -0,0 +1,57 @@
...
+ * already sorted.
+ * @param <T> type of ranges - must be a subtype {@link Comparable} as well as {@link Serializable}
+ */
+public class RangePartitioning<T extends Comparable<? super T> & Serializable> 
+implements Partitioning<T> {
+  
+  private final List<T> ranges;
+  
+  public RangePartitioning(List<T> ranges) {
+    this.ranges = ranges;
```
xitep: i would make a copy of the list here. of course, it's arguable whether it helps something or not. i'll leave it up to you.

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Sort.java
```diff
@@ -0,0 +1,305 @@
...
+  public static class ByBuilder<IN> {
+    private final String name;
+    private final Dataset<IN> input;
+
+    ByBuilder(String name, Dataset<IN> input) {
+      this.name = requireNonNull(name);
+      this.input = requireNonNull(input);
+    }
+
+    public <S extends Comparable<? super S>> WindowByBuilder<IN, S> by(UnaryFunction<IN, S> sortByFn) {
```
vanekjar: I vote for `.by`. Sounds more natural to me. It is `sort using [some] function` vs `sort by [some] field` and our meaning is more closer to the field.
Also the same is in [Spark API](https://spark.apache.org/docs/2.0.1/api/java/org/apache/spark/rdd/RDD.html#sortBy(scala.Function1,%20boolean,%20int,%20scala.math.Ordering,%20scala.reflect.ClassTag)) with `sortBy` method.

xitep: regarding myself: i'm fine with both. so `.by` is fully fine with me. just wanted to through the question onto the table.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/batch/SortTranslator.java
vanekjar: Yes, current implementation will be definitely slow as hell. We could rewrite it to do the same trick as introduced in Spark (see [ExternalSorter](https://github.com/apache/spark/blob/39e2bad6a866d27c3ca594d15e574a1da3ee84cc/core/src/main/scala/org/apache/spark/util/collection/ExternalSorter.scala#L406)), but it will be non-trivial change.
I think this can be separate issue.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/batch/SortTranslator.java
horkyada: You mean create new Wrapper for this? Because the `PartitionerWrapper` you mention is used by all other translators.

xitep: yes, exactly. i meant creating a new wrapper.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/batch/SortTranslator.java
xitep: 👍

horkyada: In Flink it should be sufficient to return the keys inside the `Tuple` implementing [NormalizableKey](https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/types/NormalizableKey.java). 
 - we can implement wrappers for euphoria-known windows (e.g. TimeWindowWrapper would return key based on endTimestamp), it ensures that records from different windows would be compared quickly
 - on the other hand, we don't know the user-defined key type and so it does not help the most common scenario just to sort the batch input on global window

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Sort.java
xitep: update: after some more investigations, `-1` should not be returned at the moment.

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/SortTranslator.java
```diff
@@ -0,0 +1,172 @@
...
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public int getPartition(Object el) {
+      Tuple3<Integer, Window, Comparable> t = (Tuple3<Integer, Window, Comparable>) el;
+      return t._1();
+    }
+  }
+  
+  private static class TripleComparator implements Comparator<Tuple3<Integer, Window, Comparable>>, Serializable {
```
xitep: i think we should strongly consider switching over to scala. all of this "triple-comparator business" should be this in scala:

```scala
import scala.math.Ordering.Implicits._
var x = (1,2,3) < (1,2,4)
```

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/batch/SortTranslator.java
xitep: the suggestion with normalized-key sounds good to me; as we discussed offline, the benefit will be questionable especially since the type of the actual sort-key is unknown. i'd like to ask you to merely set up a ticket, such that we won't forget to resolve this later.

xitep:
Thank you for the update! 👍 👍

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Sort.java
```diff
@@ -0,0 +1,345 @@
...
+ * from the input object. The extracted object is than passed to a provided partitioner
+ * to partition the result.<p>
+ * 
+ * To ensure that all elements from a specific range end up 
+ * in the same partition - i.e. perform total sort ordering - it is recommended 
+ * to use {@link RangePartitioning}, otherwise the user is responsible for his partitioning.<p>
+ * 
+ * If user does not provide custom {@link Partitioning} or {@link Partitioner} and number of input 
+ * partitions differs from 1 (single partition), the program crashes, because runtime does not 
+ * know how to partition the result. Input sampling is not supported for now.<p>
```
vanekjar: Does it really crash? What is the reason not to use default `HashPartitioner` with same level of parallelism as input dataset? The result obviously wouldn't be totally ordered but it would be equal to Flink's "sort partition" transformation.

vanekjar:
Cool! Many thanks :+1:

PR #100, opened 2017-04-20
#95 Rename Batch to GlobalWindowing

### LINKED ISSUE #95: Consider aligning naming of "batch windowing"
Apache Beam as well as Apache Flink refer to what we named `Batch` and `Batch.BatchWindow` as `GlobalWindows` and `GlobalWindow` respectively. I think the term "global" makes more sense. 1) Windowing is a separate concept/value to "batch/stream processing". 2) By aligning the names we create less confusion for new users.

In particular, I'd like to propose renaming:

* `Batch` => `GlobalWindowing`
* `Batch.BatchWindow` => `GlobalWindowing.GlobalWindow` or just `GlobalWindowing.Window`

Please comment if renaming doesn't make sense to you.

xitep:
* Resolves #95

### DIFF
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/dataset/windowing/GlobalWindowing.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/dataset/windowing/GlobalWindowing.java
@@ -53,17 +53,20 @@ public final class GlobalWindowing<T>
     }
 
     @Override
-    public int compareTo(Window o) {
+    public int compareTo(GlobalWindowing.Window o) {
       return 0;
     }
   } // ~ end of Label
 
   private final static GlobalWindowing<?> INSTANCE = new GlobalWindowing<>();
+  private final static Iterable<Window> INSTANCE_ITER =
+      Collections.singleton(Window.INSTANCE);
+
   private GlobalWindowing() {}
 
   @Override
   public Iterable<Window> assignWindowsToElement(WindowedElement<?, T> el) {
-    return Collections.singleton(Window.INSTANCE);
+    return INSTANCE_ITER;
   }
 
   @Override
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/dataset/windowing/GlobalWindowing.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/dataset/windowing/GlobalWindowing.java
@@ -25,19 +25,22 @@ import java.util.Collections;
  * Windowing with single window across the whole dataset. Suitable for
  * batch processing.
  */
-public final class Batch<T>
-    implements Windowing<T, Batch.BatchWindow> {
+public final class GlobalWindowing<T>
+    implements Windowing<T, GlobalWindowing.Window> {
 
-  public static final class BatchWindow extends Window implements Comparable<BatchWindow> {
-    static final BatchWindow INSTANCE = new BatchWindow();
+  public static final class Window
+      extends cz.seznam.euphoria.core.client.dataset.windowing.Window
+      implements Comparable<Window> {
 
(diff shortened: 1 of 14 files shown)


euphoria-core/src/main/java/cz/seznam/euphoria/core/client/dataset/windowing/GlobalWindowing.java
```diff
@@ -50,27 +53,27 @@ private Object readResolve() throws ObjectStreamException {
     }
 
     @Override
-    public int compareTo(BatchWindow o) {
+    public int compareTo(Window o) {
```
je-ik: This seems to be non conformant with the `equals` method. Is it on purpose?

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/dataset/windowing/GlobalWindowing.java
```diff
@@ -50,27 +53,27 @@ private Object readResolve() throws ObjectStreamException {
...
-  private final static Batch<?> INSTANCE = new Batch<>();
-  private Batch() {}
+  private final static GlobalWindowing<?> INSTANCE = new GlobalWindowing<>();
+  private GlobalWindowing() {}
 
   @Override
-  public Iterable<BatchWindow> assignWindowsToElement(WindowedElement<?, T> el) {
-    return Collections.singleton(BatchWindow.INSTANCE);
+  public Iterable<Window> assignWindowsToElement(WindowedElement<?, T> el) {
+    return Collections.singleton(Window.INSTANCE);
```
je-ik: Here is possible a micro optimization - we could create the set and store it in static field. It should be immutable, so it seems safe.

euphoria-inmem/src/main/java/cz/seznam/euphoria/inmem/InMemExecutor.java
```diff
@@ -712,7 +711,7 @@ private InputProvider execReduceStateByKey(
                   (oldVal, newVal) -> oldVal < newVal ? newVal : oldVal);
               // determine partition
               Object key = keyExtractor.apply(datum.getElement());
-              final Iterable<Window> targetWindows;
+              final Iterable<cz.seznam.euphoria.core.client.dataset.windowing.Window> targetWindows;
```
je-ik: Why we dropped the import here and are referencing the class by whole name?

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/dataset/windowing/GlobalWindowing.java
xitep: hm, i think it is. `Window` here refers here to `GlobalWindowing.Window`; i'll make that explicit.

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/dataset/windowing/GlobalWindowing.java
xitep: yeah, why not. seem safe to me as well.

euphoria-inmem/src/main/java/cz/seznam/euphoria/inmem/InMemExecutor.java
xitep: oops ... will fix that; thanks :+1:

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/dataset/windowing/GlobalWindowing.java
je-ik: Aj, you're right, my fault! :)

PR #101, opened 2017-04-20
Unify pair#getKey/getFirst and #getValue/getSecond

xitep:
having two distinct ways to access the same property is confusing and hurts code readability. This PR drops the historical `Map.Entry` implementation.

### DIFF
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/util/Pair.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/util/Pair.java
@@ -19,12 +19,12 @@ import java.util.Comparator;
 import java.util.Objects;
 
 /**
- * Pair of any types. The pair has to fit in memory and is read-only.
+ * A pair, i.e. a tuple of two elements.
  *
- * @param <K> the type of the key - first element of the pair
- * @param <V> the type of the value - second element of the pair
+ * @param <K> the type of the first element of the pair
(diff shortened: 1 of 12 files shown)


euphoria-core/src/main/java/cz/seznam/euphoria/core/client/util/Pair.java
```diff
@@ -19,12 +19,12 @@
 import java.util.Objects;
 
 /**
- * Pair of any types. The pair has to fit in memory and is read-only.
+ * A pair, i.e. a tuple of two elements.
```
vanekjar: I see you removed the notice about mutability. What if we make the `Pair` really mutable by adding setters? I think most of implementations would benefit from this. Now we create a lot of short-term living objects because of `Pair` immutability. Is there any reason not go this way?
But maybe this a topic for separate issue.

xitep: ah, thanks for raising the topic about the pair's mutability! (the removal of this information was actually not intentional.) i fully agree, sooner or later we will have to think about the mutability to keep GC as low as possible. in general, it makes sense to have these types mutable; user provided types may be mutable as well and there's nothing special about the tuple classes - at least i don't see it.

however, to properly leverage the advantages of mutable tuples, we'll also have to think about object re-use in general within euphoria. this can be pretty far reaching. i'd say we tackle this topic as a separate issue.

PR #108, opened 2017-04-27
#! Remove obsolete classes and forbid custom Pairs/Triples

xitep:
* Seem to be some left over from earlier times
* Actually replaced by `StateAwareWindowWiseSingleInputOperator`


vanekjar:
Thanks for the cleanup :+1:

PR #109, opened 2017-04-27
Javadoc refactoring

horkyada:
Added interfaces (in package `cz.seznam.euphoria.core.client.operator.builder`) with general javadocs that are then used in operator builders to inherit these javadocs.

Some (probably not all) javadocs added to most of operators.


xitep:
i fear this will require `Builders` to be public in the end :/

xitep:
Thank you so much! 💯

PR #112, opened 2017-04-28
#13 [euphoria-flink] Batch executor uses hash code as a key in shufflles

vanekjar:
I have finally finished this improvement. Comparing elements in shuffle by hashcode (integer) is much faster than deserializing and calling `.compareTo` on original instance.

The result is batch executor's performance is a about 3 times better compared to previous version.

### DIFF
--- a/euphoria-flink/src/main/java/cz/seznam/euphoria/flink/batch/ReduceByKeyTranslator.java
+++ b/euphoria-flink/src/main/java/cz/seznam/euphoria/flink/batch/ReduceByKeyTranslator.java
@@ -149,12 +149,11 @@ public class ReduceByKeyTranslator implements BatchOperatorTranslator<ReduceByKe
   static class RBKKeySelector
           implements KeySelector<BatchElement<Window, Pair>, Integer> {
 
-    private final Tuple2 tuple = new Tuple2();
-    
     @Override
     public Integer getKey(
             BatchElement<Window, Pair> value) {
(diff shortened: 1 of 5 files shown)


euphoria-flink/src/main/java/cz/seznam/euphoria/flink/batch/ReduceByKeyTranslator.java
```diff
@@ -159,16 +172,64 @@ public DataSet translate(FlinkOperator<ReduceByKey> operator,
...
+                       org.apache.flink.util.Collector<BatchElement<Window, Pair>> out) {
+
+      // Tuple2[Key, Window] => Reduced Value
+      Map<Tuple2, TimestampedElement> reducedValues = new HashMap<>();
+
+      for (BatchElement<Window, Pair> batchElement : values) {
+        Object key = batchElement.getElement().getFirst();
+        Window window = batchElement.getWindow();
+
+        Tuple2 kw = new Tuple2<>(key, window);
```
xitep: may i ask for a short comment here explaining why the order is reversed compared to the order of the two items in `RBKKeySelector`; i know its in purpose; but i fear i'll forget about it in half a year.

vanekjar: Sure :+1:

PR #114, opened 2017-05-03
Early emitting state

### LINKED ISSUE #105: euphoria-core: reduce-state-by-key: merging windowing and emitting state earlier than flush
`StateFactory` is invoked with an (output) context to allow the corresponding states to emit results as early as possible. however, together with a merging windowing strategy, e.g. `Session`, this can lead to wrong results, i.e. leading to results ending up in wrong windows. Currently known to suffer under this issue is the `JoinState` implementation. This issue is similar to #43.

xitep:
* Fixes the Join operator to produce stable output, i.e. without early emitting
* Provides an experimental early emitting implementation of the JoinState, similar to the previously existing implementation
* Support early emitting on the inmem and flink streaming executors through an executor parameter
* Resolves #105

### DIFF
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Join.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Join.java
@@ -390,8 +390,9 @@ public class Join<LEFT, RIGHT, KEY, OUT, W extends Window>
 
   /**
    * An implementation of the join state which produces results, i.e. emits
-   * output, as soon as possible. It has at least the following short commings
-   * and should be used with care:
+   * output, as soon as possible. It has at least the following short comings
+   * and should be used with care (see https://github.com/seznam/euphoria/issues/118
+   * for more information):
    *
    * <ul>
    *   <li>This implementation will break the join operator if used with a
@@ -405,6 +406,7 @@ public class Join<LEFT, RIGHT, KEY, OUT, W extends Window>
    *        thus, marking earlier - but actually still not too late time-sliding
    *        windows as late comers.</li>
    * </ul>
+   *
    *
    */
   @Experimental
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Join.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Join.java
@@ -400,6 +400,10 @@ public class Join<LEFT, RIGHT, KEY, OUT, W extends Window>
    *   <li>This implementation cannot be used together with early triggering
    *        on any windowing strategy as it will emit each identified pair
    *        only once during the whole course of the state's life cycle.</li>
+   *   <li>This implementation will also break time-sliding windowing, as
+   *        it will raise the watermark too quickly in downstream operators,
+   *        thus, marking earlier - but actually still not too late time-sliding
+   *        windows as late comers.</li>
    * </ul>
    *
    */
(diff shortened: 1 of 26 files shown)


euphoria-flink/src/main/java/cz/seznam/euphoria/flink/batch/BatchStateStorageProvider.java
```diff
@@ -140,22 +141,34 @@ public void add(T element) {
 
     @Override
     public Iterable<T> get() {
+      return this::iter;
+    }
 
+    private Iterator<T> iter() {
+      FileInputStream finput;
```
vanekjar: Maybe it is not necessary to store `finput` explicitly just to call close on it. Now the `FileInputStream` is actually closed twice. Once by `Closeables.closeQuietly(finput)` and second transitively from `input.close()`.
But it does not cause any harm.

xitep: true! will throw it out. thanks 👍

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Join.java
```diff
@@ -436,19 +514,21 @@ public boolean isOuter() {
...
-              partitioning
-        );
+        getName() + "::ReduceStateByKey",
+        flow,
+        union.output(),
+        keyExtractor,
+        e -> e,
+        getWindowing(),
+        getEventTimeAssigner(),
+        (StorageProvider storages, Context ctx) ->
```
je-ik: Could we make this `Context` here `Optional` instead of passing `null`? This is part of client API, ale I think it would be a little cleaner.

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/state/StateFactory.java
```diff
@@ -25,6 +27,17 @@
...
+   *
+   * @param storageProvider the provider for physical storage accessors
+   * @param context a context allowing the newly created state for the
+   *                duration of its existence to emit output elements
+   *
+   * @return a newly created state
+   */
+  STATE createState(StorageProvider storageProvider,
+                    @Experimental("may break merging windowing and downstream watermarking")
+                    @Nullable Context<OUT> context);
```
je-ik: OK, this seems documented enough, so we can keep it this way. :)

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Join.java
```diff
@@ -307,99 +305,179 @@ public static OfBuilder named(String name) {
...
+   *   <li>This implementation will break the join operator if used with a
+   *        merging windowing strategy, since items will be emitted under the
+   *        hood of a non-final window.</li>
+   *   <li>This implementation cannot be used together with early triggering
+   *        on any windowing strategy as it will emit each identified pair
+   *        only once during the whole course of the state's life cycle.</li>
+   *   <li>This implementation will also break time-sliding windowing, as
+   *        it will raise the watermark too quickly in downstream operators,
+   *        thus, marking earlier - but actually still not too late time-sliding
+   *        windows as late comers.</li>
```
xitep: yes, you exactly hit the point regarding the watermark issue! i will have to investigate around this as i fear it'll break attached windowing. anyway i would do such a change separately and will set up a ticket for that.

je-ik: Yes, this is definitely a separate issue, we might only link the issue here with comment, so that we know that when we resolve the issue, we don't leave the comment here. We can discuss the implications on attached windowing in the new ticket.

PR #117, opened 2017-05-05
 #! polish: removed unnecessary null check, variable reuse

je-ik:


euphoria-inmem/src/main/java/cz/seznam/euphoria/inmem/InMemExecutor.java
```diff
@@ -686,9 +686,7 @@ private InputProvider execReduceStateByKey(
             Datum.watermark(maxElementTimestamp.get()),
             ret, readerId, clocks, true);
       };
-      if (emitStrategies != null) {
```
je-ik: This check is not needed, the variable is never null.

xitep:
👍 merci! will merge as soon as travis gives green light.

PR #126, opened 2017-05-12
#121 allow multiple elements from reduce by key

### LINKED ISSUE #121: Feature: allow multiple elements to be emitted from ReduceByKey
Currently, `ReduceByKey` supports only single element to be emitted from the reduce function. We should relax this to support multiple emitted values by adding `reduceBy(UnaryFunctor<Iterable<T>, OUT>)`. The original `reduceBy` will then be just a special case with exactly single emitted value.

je-ik:
closes #121

### DIFF
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/ReduceByKey.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/ReduceByKey.java
@@ -123,13 +123,14 @@ public class ReduceByKey<IN, KEY, VALUE, OUT, W extends Window>
 
 
     /**
-     * Define a function that reduces all values related to one key into one result object.
+     * Define a function that reduces all values related to one key into one or more
+     * result objects.
      * The function is not combinable - i.e. partial results cannot be made up before shuffle.
      * To get better performance use {@link #combineBy} method.
      *
      * @param <OUT> type of output element
      *
-     * @param reducer function that reduces all values into one output object
+     * @param reducer function that reduces all values into output values
      *
      * @return next builder to complete the setup of the {@link ReduceByKey} operator
      */
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/ReduceByKey.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/ReduceByKey.java
@@ -36,6 +36,7 @@ import cz.seznam.euphoria.core.client.operator.state.StorageProvider;
 import cz.seznam.euphoria.core.client.operator.state.ValueStorage;
 import cz.seznam.euphoria.core.client.operator.state.ValueStorageDescriptor;
 import cz.seznam.euphoria.core.client.util.Pair;
+import cz.seznam.euphoria.core.executor.util.SingleValueContext;
 
 import javax.annotation.Nullable;
 import java.util.Arrays;
@@ -394,28 +395,6 @@ public class ReduceByKey<IN, KEY, VALUE, OUT, W extends Window>
   }
 
 
-  public static class SingleValueContext<T> implements Context<T> {
-
-    T value;
-
-    @Override
-    public void collect(T elem) {
-      value = elem;
-    }
-
-    @Override
-    public Object getWindow() {
-      throw new UnsupportedOperationException(
-          "The window is unknown when applying combinable function");
-    }
-
-    public T getValue() {
-      return value;
-    }
-
-  };
-
-
   static class CombiningReduceState<E>
           implements State<E, E>, StateSupport.MergeFrom<CombiningReduceState<E>> {
 
@@ -457,7 +436,7 @@ public class ReduceByKey<IN, KEY, VALUE, OUT, W extends Window>
         this.storage.set(element);
       } else {
         this.reducer.apply(Arrays.asList(v, element), context);
-        this.storage.set(context.getValue());
+        this.storage.set(context.getAndResetValue());
       }
     }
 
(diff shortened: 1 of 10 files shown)


je-ik:
Hi guys, this is ready for review for me. Thanks for any comments.

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/ReduceByKey.java
```diff
@@ -352,19 +391,58 @@ public boolean isCombinable() {
...
+
+    @Override
+    public void collect(T elem) {
+      value = elem;
+    }
+
+    @Override
+    public Object getWindow() {
+      throw new UnsupportedOperationException(
+          "The window is unknown when applying combinable function");
```
xitep: * i don't think this is generally true. i suppose the difficulty here is "merely" getting at the corresponding context information during the reduction through the state. and it indicates to me that the (api) boundary between the operator and the state implementation is currently insufficient. (this memory leak appears to be repeated also in the spark and flink RBK translators.)
* further, i don't think exposing this `SingleValueContext` is a good idea (it seems to me like an impl detail of `CombiningReduceState`).

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/ReduceByKey.java
```diff
@@ -391,7 +470,8 @@ public void add(E element) {
       if (v == null) {
         this.storage.set(element);
       } else {
-        this.storage.set(this.reducer.apply(Arrays.asList(v, element)));
+        this.reducer.apply(Arrays.asList(v, element), context);
+        this.storage.set(context.getValue());
```
xitep: the `context` should probably be cleared at this point. it seems there is no reason to further hard reference the "value" after storing it in the backend storage. it could be GC'd right away.

euphoria-operator-testkit/src/main/java/cz/seznam/euphoria/operator/test/ReduceByKeyTest.java
```diff
@@ -86,7 +88,7 @@ public void testReductionType0() {
         return ReduceByKey.of(input)
             .keyBy(e -> e % 2)
             .valueBy(e -> e)
-            .reduceBy(Sets::newHashSet)
+            .reduceBy((ReduceFunction<Integer, HashSet<Integer>>) Sets::newHashSet)
```
xitep: :(

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/ReduceByKey.java
je-ik: What this raelly ment to say is - you should never see this exception. :)
Currently, you cannot get to the context in combinable RBK, because you have to pass in `CombinableReduceFunction` which has no access to it. This exception will be therefore never thrown in user code.
The problem with this in general is when applying combinable reduce function to merging windowing, then the window is trully unknown and there is no way around this. It isn't a real problem, because the API prevents you from accessing the window when applying combinable function, as I metioned above.

je-ik: The second point is about exposing the `SingleValueContext` - I don't like to expose it on RBK either, but I didn't find a better place to put it and not to reimplement it inside each executor. You are right that it could be put inside the `CombiningReduceState`, but then it would have to be reimplemented inside `euphoria-flink` and `euphoria-spark` appropriately. Maybe I could move this into `cz.seznam.euphoria.core.executor.util` package?

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/ReduceByKey.java
je-ik: Okay, I can change the contract to that instead of `getValue()` a will introduce `getAndResetValue()` which will do as it name suggests. :)

euphoria-operator-testkit/src/main/java/cz/seznam/euphoria/operator/test/ReduceByKeyTest.java
je-ik: Yeah, there is no way around this (that I can think of). But I don't think that this will make any real user-facing problem, most of the time, the lambda will be correctly disambiguated.

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/ReduceByKey.java
xitep: * regarding the exception: that's true! i overlooked that. (assuming users never implement their own `ReduceFunctor` with an overridden `isCombinable`.) but then i don't really like it :( but i can live with it :) no need to change it. btw: is there a way for the executor to grasp the `CombinableReduceFunction` directly and be based directly on that?
* i'm okey with the duplication and also okey with the `util` package.

je-ik: Yes, you are right that users could implement their own `ReduceFunctor` with `isCombinable` returning `true`. I think there was some other issue with returning the window, I will lokk at it again and we can discuss it further.

je-ik: Regarding the other question - if the executor can work directly with `CombinableReduceFunction` - the answer is obvious, yes it can, at the expense of having two functional fields, one for non combinable and one for combinable reducing function. I think this would result in much more messy situation then we have here and I would prefer to avoid it.

Do we have any other options to solve this problem?

je-ik: I just realized the original problem of combinable `ReduceFunctor` to have access to the reduced window - if the windowing is merging windowing, then the window label will change during reducing (i.e. some elements will be reduced as if the window was X, while some other would be reduced as if the window would be Y, and the results will be randomly mixed together). I think it is better not to allow user code to access the window by throwing the `UnsupportedOperationException`. We could make this dependent on the merging windowing, but I think that a more straightforward approach is just to disable it all at once to stick to the least astonishment principle.

je-ik:
Hi @xitep, I incorporated the changes we agreed upon. Any more feedback would be appreciated. Thanks! :)

xitep:
looks good to me now. thanks for the updates! 👍 feel free to merge into `master` if it is final from your side.

PR #127, opened 2017-05-12
Rework event time assignment

### LINKED ISSUE #119: euphoria-flink: late comers in join
This test consistently fails on the flink streaming executor:

```java
  // ~ all of the inputs fall into the same session window (on the same key)
  // ~ we expect the result to reflect this fact
  // ~ note: no early triggering
  @Processing(Processing.Type.UNBOUNDED)
  @Test
  public void innerJoinOnSessionWindowingNoEarlyTriggering() {
    execute(new JoinTestCase<
        Pair<String, Long>,
        Pair<String, Long>,
        Triple<TimeInterval, String, String>>() {

      @Override
      protected Partitions<Pair<String, Long>> getLeftInput() {
        return Partitions.add(
            Pair.of("fi", 1L),
            Pair.of("fa", 2L))
            .build();
      }

      @Override
      protected Partitions<Pair<String, Long>> getRightInput() {
        return Partitions.add(
            Pair.of("ha", 1L),
            Pair.of("ho", 4L))
            .build(Duration.ofMillis(10000), Duration.ofMillis(100));
      }

      @Override
      public int getNumOutputPartitions() {
        return 1;
      }

      @Override
      protected Dataset<Triple<TimeInterval, String, String>>
      getOutput(Dataset<Pair<String, Long>> left, Dataset<Pair<String, Long>> right) {
        Dataset<Pair<String, Triple<TimeInterval, String, String>>> joined =
            Join.of(left, right)
                .by(p -> "", p -> "")
                .using((Pair<String, Long> l, Pair<String, Long> r, Context<Triple<TimeInterval, String, String>> c) ->
                    c.collect(Triple.of((TimeInterval) c.getWindow(), l.getFirst(), r.getFirst())))
                .windowBy(Session.of(Duration.ofMillis(10)), Pair::getSecond, Pair::getSecond)
                .setNumPartitions(1)
                .output();
        return MapElements.of(joined).using(Pair::getSecond).output();
      }

      @Override
      public void validate(Partitions<Triple<TimeInterval, String, String>> partitions) {
        TimeInterval expectedWindow = new TimeInterval(1, 14);
        assertUnorderedEquals(
            Arrays.asList(
                Triple.of(expectedWindow, "fi", "ha"),
                Triple.of(expectedWindow, "fi", "ho"),
                Triple.of(expectedWindow, "fa", "ha"),
                Triple.of(expectedWindow, "fa", "ho")),
            partitions.get(0));
      }
    });
  }
```

the point is that the right input of the join is very slow compared to the left input; the test fails due to the first element from the right input being discarded as a late comer. this is not the expected behaviour.

a quick brain-storming session and an experiment confirm, that when event time assignment happens _before_ the underlying union operation, the test passes.

xitep:
* Resolves #119

### DIFF
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/flow/Flow.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/flow/Flow.java
@@ -20,6 +20,8 @@ import cz.seznam.euphoria.core.client.dataset.Datasets;
 import cz.seznam.euphoria.core.client.io.DataSink;
 import cz.seznam.euphoria.core.client.io.DataSource;
 import cz.seznam.euphoria.core.client.io.IORegistry;
+import cz.seznam.euphoria.core.client.operator.AssignEventTime;
+import cz.seznam.euphoria.core.client.operator.ExtractEventTime;
 import cz.seznam.euphoria.core.client.operator.Operator;
 import cz.seznam.euphoria.core.util.Settings;
 import org.slf4j.Logger;
@@ -38,6 +40,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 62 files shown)


euphoria-core/src/main/java/cz/seznam/euphoria/core/client/flow/Flow.java
```diff
@@ -335,6 +338,23 @@ public Settings getSettings() {
...
+   *
+   * @param <T> the type of elements of the created input data set
+   *
+   * @param source the data source to represent as a data set
+   * @param evtTimeFn the user defined event time extraction function
+   *
+   * @return a data set representing the specified source of data with assigned
+   *          event time assigned
+   */
+  public <T> Dataset<T> createInput(DataSource<T> source, ExtractEventTime<T> evtTimeFn) {
```
je-ik: :+1: this is really cool. When we discussed this I didn't quite realized that this form of time assignment is just a special case of the one we agreed to implement.

euphoria-flink/src/test/java/cz/seznam/euphoria/flink/streaming/RBKTimeWindowTest.java
```diff
@@ -188,17 +189,15 @@ public void testEventWindowingNonCombining() throws Exception {
 
     Flow f = Flow.create("test-attached-windowing");
     Dataset<Pair<String, String>> reduced =
-        ReduceByKey.of(f.createInput(source))
+        ReduceByKey.of(f.createInput(source, Triple::getThird))
```
je-ik: Yeah, here we can see how it help readability a lot! :+1:

PR #130, opened 2017-05-17
#113 Spilling list storage

### LINKED ISSUE #113: euphoria-spark: non-heap storage provider
Spark executor's `SparkStorageProvider` keeps all states in memory. This is problematic and will sooner or later definitely hit a wall.

xitep:
* Resolves #113

### DIFF
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/executor/storage/FsSpillingListStorage.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/executor/storage/FsSpillingListStorage.java
@@ -32,9 +32,53 @@ import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Objects;
 
+/**
+ * A list storage implementation keeping a given maximum number of received
+ * elements, eventually starting to spill them to the local file system.
+ * Intended for use in batch executors.
+ *
+ * @param <T> the type of elements stored in this list storage
+ */
 public class FsSpillingListStorage<T> implements ListStorage<T> {
 
+  /**
+   * A factory for spill files.
+   */
+  @FunctionalInterface
+  public interface SpillFileFactory {
+    /**
+     * Invoked to request a unique path to a new spill file.
+     *
+     * @return the path to a new spill file
+     */
+    File newSpillFile();
+  }
+
+  /**
+   * The default spill file factory to create new files in the current
+   * working directory.
+   */
+  public static class DefaultSpillFileFactory implements SpillFileFactory {
+    static final SpillFileFactory INSTANCE = new DefaultSpillFileFactory();
+    private DefaultSpillFileFactory() {}
+
+    public static SpillFileFactory getInstance() {
+      return INSTANCE;
+    }
+
+    @Override
+    public File newSpillFile() {
+      try {
+        Path workDir = FileSystems.getFileSystem(new URI("file:///")).getPath("./");
+        return Files.createTempFile(workDir, getClass().getName(), ".dat").toFile();
+      } catch (Exception e) {
+        throw new IllegalStateException(e);
+      }
+    }
+  }
+
   private final SerializerFactory serializerFactory;
+  private final SpillFileFactory spillFileFactory;
   private final int maxElemsInMemory;
 
   // ~ new elements are appended to this list and eventually this list
@@ -47,8 +91,11 @@ public class FsSpillingListStorage<T> implements ListStorage<T> {
   private SerializerFactory.Serializer.OutputStream serializerStream;
   private boolean needsFlush;
 
-  public FsSpillingListStorage(SerializerFactory serializerFactory, int maxElemsInMemory) {
+  public FsSpillingListStorage(SerializerFactory serializerFactory,
+                               SpillFileFactory spillFileFactory,
+                               int maxElemsInMemory) {
     this.serializerFactory = Objects.requireNonNull(serializerFactory);
+    this.spillFileFactory = Objects.requireNonNull(spillFileFactory);
     this.maxElemsInMemory = maxElemsInMemory;
   }
 
@@ -87,7 +134,9 @@ public class FsSpillingListStorage<T> implements ListStorage<T> {
         try {
           is = serializerInstance.newInputStream(new FileInputStream(storageFile));
         } catch (FileNotFoundException e) {
-          throw new IllegalStateException("Failed to open spilling storage: " + storageFile, e);
+          throw new IllegalStateException(
+              "Failed to open spilling storage: "
+                  + storageFile, e);
         }
       } else {
         is = null;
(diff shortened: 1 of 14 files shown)


xitep:
a note for myself: once merged, don't forget to update [the configuration page in the wiki](https://github.com/seznam/euphoria/wiki/Configuration)

euphoria-core/src/main/java/cz/seznam/euphoria/core/executor/storage/FsSpillingListStorage.java
```diff
@@ -0,0 +1,148 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
+public class FsSpillingListStorage<T> implements ListStorage<T> {
```
je-ik: It would be cool to add some basic description, mostly when to use this storage.

euphoria-core/src/main/java/cz/seznam/euphoria/core/executor/storage/FsSpillingListStorage.java
```diff
@@ -0,0 +1,148 @@
...
+      }
+    }
+  }
+
+  private void initDiskStorage() {
+    assert storageFile == null;
+    assert serializerInstance == null;
+    assert serializerStream == null;
+    try {
+      Path workDir = FileSystems.getFileSystem(new URI("file:///")).getPath("./");
```
je-ik: I think it would be useful if the scheme was not hard coded here. It can default to `file://`, but would it be possible let the user specify some other scheme (e.g. `hdfs://`)?

euphoria-core/src/main/java/cz/seznam/euphoria/core/executor/storage/SerializerFactory.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package cz.seznam.euphoria.core.executor.storage;
+
+import java.io.Serializable;
+
+public interface SerializerFactory extends Serializable {
+
+  interface Serializer {
+
+    interface OutputStream {
```
je-ik: This should maybe extend `AutoCloseable` or `Closeable`?

euphoria-operator-testkit/src/main/java/cz/seznam/euphoria/operator/test/ReduceByKeyTest.java
```diff
@@ -126,119 +126,119 @@ public void validate(Partitions<Pair<Integer, HashSet<Integer>>> partitions) {
...
+//        List<Pair<Integer, Long>> second = partitions.get(1);
+//        assertUnorderedEquals(Arrays.asList(
+//            Pair.of(1, 1L), Pair.of(3, 2L),  // first window
+//            Pair.of(1, 1L), Pair.of(3, 2L),  // second window
+//            Pair.of(1, 2L), Pair.of(3, 2L),  // third window
+//            Pair.of(3, 1L)),                 // fourth window
+//            second);
+//      }
+//    });
+//  }
```
je-ik: Why do we comment this out? If it is obsolete for some reason, shouldn't we just drop it?

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/SparkSerializerFactory.java
```diff
@@ -0,0 +1,155 @@
...
+            return n;
+          }
+          if (streamFinished) {
+            throw new NoSuchElementException("End of file reached!");
+          }
+          tryReadNext();
+        }
+      }
+
+      @SuppressWarnings("BC_IMPOSSIBLE_INSTANCEOF")
```
je-ik: I think this was renamed to `SuppressFBWarnings` to avoid collision with the `java.lang` version. Can we also add a little note about why we suppress the warning?

euphoria-core/src/main/java/cz/seznam/euphoria/core/executor/storage/FsSpillingListStorage.java
```diff
@@ -0,0 +1,148 @@
...
+      storageFile = Files.createTempFile(workDir, getClass().getName(), ".dat").toFile();
```
je-ik: Maybe the directory where to put this file should be configurable? If I read this correctly, it will create temp file in working directory (which could be default), but it might be useful to specify some other directory (or even several directories). If several directories would be used, then it could have performance benefits over using just a single one, if we could actually open file on each of them and then somehow (maybe completely at random) spill elements into multiple files (and therefore potentially across multiple disks).

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/SparkSerializerFactory.java
xitep: uf ... that's a left over. thanks for the catch! 👍

euphoria-operator-testkit/src/main/java/cz/seznam/euphoria/operator/test/ReduceByKeyTest.java
xitep: oops ... thanks!

euphoria-core/src/main/java/cz/seznam/euphoria/core/executor/storage/SerializerFactory.java
xitep: yes, that's good idea 👍

euphoria-core/src/main/java/cz/seznam/euphoria/core/executor/storage/FsSpillingListStorage.java
xitep: in fact, i was already considering to "outsource" this piece into the executors themself. i think i will do so now, executors are then free to choose what they prefer. thanks for the push 👍

euphoria-core/src/main/java/cz/seznam/euphoria/core/executor/storage/FsSpillingListStorage.java
xitep: will do. thanks!

euphoria-core/src/main/java/cz/seznam/euphoria/core/executor/storage/SerializerFactory.java
xitep: hm ... this imposes difficulty without bringing in a lot :/

je-ik: I think you can redeclared the `close` method not to throw exceptions.

xitep: Indeed! :) Gracias!

xitep:
Once again thanks for your review, guys. I just made some updates hopefully fixing all of the previous troubles.

PR #132, opened 2017-05-23
#31 [euphoria-core] Implementation of accumulator API

### LINKED ISSUE #31: euphoria-core: consider adding support for metrics
Client programs often have the need to gather and report metrics, e.g. counters, timers, gauges. The most obvious use-cases are debugging and monitoring. Currently, euphoria itself lacks any client side support in this regard. Actually, it's a question whether it makes sense to have native support through euphoria for metrics/aggregators or whether to leave it to 3rd-party APIs to take over here.

A typically example is to measure the execution time of a lengthy operation, like in the following example:

```java
Dataset<String> words =
    FlatMap.of(lines)
    .using((String line, Context<String> context) -> {

      // 1)
      long start = System.nanoTime();
      // do some expensive work here
      context.timer("foobar-duration").record(System.nanoTime() - start, TimeUnit.NANOTIME);

     // 2)
     if (someCondition) {
       context.counter("someCondition-true").increment();
     }

     // 3)
     start = System.nanoTime();
     // do some more expensive work
     context.timer("quux-duration").record(System.nanoTime() - start, TimeUnit.NANTIME);
    })
    .output();
```

This example suggests, that euphoria's `Context` object provides a metrics API. However, to make this globally practical, it precludes that such a context object be available to most user defined functions, which is not the case at the moment.

If we end up supporting a client side API for metrics, we'll need to think about about an SPI at the same time, such that different metric reporting backends/implementation can be hooked in independently of the actual flow execution engine.

vanekjar:
I implemented a basic accumulator API under `euphoria-core` as well with a sample implementation in `euphoria-flink` executor. Other implementations will come later.

There are three fixed types of accumulators:
 * Counter - Basic Hadoop style integral counter
 * Histogram - Accumulator collecting distribution of values
 * Timer - Collects distribution of time measurements (very similar to Histogram, extended by time units)

From user function perspective accumulators can be used in the following way:

```java
// MapElements without counters, uses UnaryFunction lambda
Dataset<String> words = MapElements.of(dataset)
    .using(s -> s.toLowerCase())
   
// MapElements with access to context, uses UnaryFunctionEnv lambda.
// Obtains read-only context (without access to .collect() method) as an argument.
Dataset<String> mapped = MapElements.of(dataset)
    .using((input, context) -> {
      // use simple counter
      context.getCounter("my-counter").increment();
  
      // use histogram
      context.getHistogram("my-hist").add(5L);

      // use timer
      context.getTimer("my-timer").add(Duration.ofSeconds(5L));
  
      return input.toLowerCase();
    })
```

When creating an `executor` user can define preferred accumulator backend:

```java
FlinkExecutor executor = new FlinkExecutor();
executor.setAccumulatorProvider(FlinkNativeAccumulators.getFactory());
// or 
executor.setAccumulatorProvider(VoidAccumulatorProvider.getFactory());
```

-----

I appreciate a thorough review of the suggested core API as well as any feedback.

Thanks!

closes #31


je-ik:
Hi @vanekjar ,
the code snippet you sent looks really good. I like the possibility to specify the accumulator backend, this seems really conceptual to me. But - what will you do with the accumulators? How do you access them - in both streaming and batch scenarios? I think the only purpose of the accumulators is not only to display them on internal web UI of a specific executor, or is it?

vanekjar:
Hi @je-ik, many thanks for the feedback. You actually brought up a great topic to be discussed.

I haven't actually decided yet how to access accumulators by general API. It seems it should be dependent on a particular `AccumulatorProvider`. It means accumulators won't be accessible via executor API itself. 

For example `FlinkNativeAccumulators` provider could expose accumulator results via it's own API like `Map<String, Object> FlinkNativeAccumulators.getAllAccumulators()`. But if we agree on that accumulators shouldn't drive business logic it's questionable if it's needed. Besides that there is a Flink REST API providing similar data to Flink UI web including accumulators.

Anyway if we're talking about other providers serving as remote collectors (_Splunk_ or _Telegraf_) there is no need for the accumulator results to be accessed from the code. The results are then presented somewhere else in the 3rd party tool (_Grafana_).

je-ik:
@vanekjar could you provide some code example about how you imagine this should work? On a first thought, it seems odd to me to provide API for creating accumulators and not to provide API to access them. Let's say I would like to store the counters inside driver and provide them as JMX for access by any other tool. I'd like to update the JMX say each 10 seconds. Can I solve this?

vanekjar:
I am afraid there is no general solution for that on Euphoria layer level. Since the driver abilities are bounded by the used framework (Spark, Flink) it won't be easy to collect the accumulators in a standardized way.

It would be easy to provide JMX interface for each task manager (JVM), but these results are just partial.  To provide the aggregated results the easiest way (on Flink) would be to monitor the Flink REST API and expose the results via JMX if needed.
https://ci.apache.org/projects/flink/flink-docs-release-1.2/monitoring/rest_api.html

Anyway if there is a general way how to collect results in driver it is still possible to implement `AccumulatorProvider` into `euphoria-core` that can be used for that purpose. The current client API should be sufficient for that.

je-ik:
I'm not sure if I understood your remark about `AccumulatorProvider`, can you explain it a little more in detail?

vanekjar:
I meant you can implement your own

```java
class JMXAccumulatorProvider implements AccumulatorProvider
```

with the custom logic that will collect the accumulator values "somewhere" that you can access them in the driver and publish via JMX. If this provider is executor independent it could be part of `euphoria-core`.

je-ik:
I'm confused, doesn't the `AccumulatorProvider` have to be executor-specific? You cannot implement it in an executor-independent way, or can you?

vanekjar:
`AccumulatorProvider` is generally executor independent and it's part of the `euphoria-core`. Some providers are executor dependent such as `FlinkNativeAccumulators` or planned `SparkNativeAccumulators`, but some of them will be independent (for example in case they're collecting value to the outside systems via network).

je-ik:
OK, I understand your point now, but I'm afraid that will not be generally working without cooperation with the executor. This is because only executors are aware of lifecycle of the individual "tasks". In batch jobs, these tasks are individual JVM performing tasks that form a 'stage' (e.g. map or reduce in M/R terms), and these tasks might fail independently of the process of sending metrics. In such case, the accumulated data would be incorrect (because the data from failed task might have already been accumulated). The same holds for streaming processors, where accumulators collected between checkpoint of a state and failure of a task will be calculated multiple times.

I'm not sure how counters are implemented in spark 2.0, but before that, counters were source of misleasing data. It is nicely explained [here]([http://imranrashid.com/posts/Spark-Accumulators/).

I think this way, we could very easily go in the same direction, that is criticized in the article above, we should avoid that, is possible.

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/accumulators/Accumulator.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package cz.seznam.euphoria.core.client.accumulators;
+
+/**
+ * Accumulators collect values from user functions.
+ * Accumulators allow user to calculate statistics during the flow execution.
+ * <p>
+ * Accumulators are inspired by the Hadoop/MapReduce counters.
+ */
+public interface Accumulator {
```
je-ik: Do we need this labelling interface? Or if yes, could we move some methods here? Maybe we could unify the `add` and `increment` methods or subinterfaces and put it here.

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/functional/UnaryFunctionEnv.java
```diff
@@ -0,0 +1,41 @@
...
+import java.io.Serializable;
+
+/**
+ * Function of single argument with access to Euphoria environment via context.
+ *
+ * @param <IN> the type of the element processed
+ * @param <OUT> the type of the result applying element to the function
+ */
+@FunctionalInterface
+public interface UnaryFunctionEnv<IN, OUT> extends Serializable {
```
je-ik: I'd suggest naming this interface something like `RichUnaryFunction`, because in conjunction with @136, we could add additional methods (with default implementation) here.

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/io/Collector.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package cz.seznam.euphoria.core.client.io;
+
+/**
+ * Extends {@link Environment} with write capability. Used in
+ * user defined functors.
+ *
+ * @param <T> the type of elements collected through this context
+ */
+public interface Collector<T> extends Environment {
```
je-ik: I have a feeling this should have some other name, because besides _collecting_, it also provides access to window and possibly other stuff.

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/io/Context.java
```diff
@@ -16,27 +16,8 @@
...
-  /**
-   * Retrieves the window - if any - underlying the current
-   * execution of this context.
-   *
-   * @return {@code null} if this context is not executed within a
-   *          windowing strategy, otherwise the current window of
-   *          this context
-   */
-  Object getWindow();
+public interface Context extends Environment {
```
je-ik: What is the difference between `Context` and `Environment` then?

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/MapElements.java
```diff
@@ -136,6 +165,41 @@ public static OfBuilder named(String name) {
     return DAG.of(
         // do not use the client API here, because it modifies the Flow!
         new FlatMap<IN, OUT>(getName(), getFlow(), input,
-            (i, c) -> c.collect(mapper.apply(i)), null));
+            (i, c) -> c.collect(mapper.apply(i, new CollectorAdapter(c))), null));
```
je-ik: I think we can safely just cast the `Collector` to `Context` here, because if the user casts it back, then he knows he does something that is wrong.

euphoria-operator-testkit/src/main/java/cz/seznam/euphoria/operator/test/MapElementsTest.java
```diff
@@ -39,7 +40,7 @@ public void testOnTwoPartitions() {
       @Override
       protected Dataset<String> getOutput(Dataset<Integer> input) {
         return MapElements.of(input)
-            .using(String::valueOf)
+            .using((UnaryFunction<Integer, String>) String::valueOf)
```
vanekjar: This is related specifically to `String#valueOf` which is heavily overloaded and offers different number of arguments. It shouldn't cause any problem in a general usage.

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/accumulators/AccumulatorProvider.java
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Creates a new instance of {@link AccumulatorProvider}
+   * initialized by given settings.
+   * <p>
+   * It is required this factory is thread-safe.
+   */
+  @FunctionalInterface
+  interface Factory extends Serializable {
+    AccumulatorProvider create(Settings settings);
```
je-ik: I think the factory should have access to the executor also. Maybe we could add it as another parameter.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/accumulators/FlinkAccumulatorFactory.java
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Creates a new instance of {@link AccumulatorProvider}
+   * initialized by given settings.
+   *
+   * @param settings Euphoria settings.
+   * @param context  Flink runtime context.
+   * @return Instance of accumulator provider.
+   */
+  AccumulatorProvider create(Settings settings, RuntimeContext context);
```
je-ik: This interface probably will not be needed if the `AccumulatorProvider.Factory#create` takes the `Executor` as argument.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/accumulators/FlinkNativeAccumulators.java
```diff
@@ -0,0 +1,175 @@
...
+        System.out.println("Creating counter: " + name);
+        // register a new instance
+        acc = clz.getConstructor().newInstance();
+        context.addAccumulator(name, (org.apache.flink.api.common.accumulators.Accumulator) acc);
+      } catch (Exception e) {
+        throw new RuntimeException("Exception during accumulator initialization: " + clz, e);
+      }
+    }
+
+    if (!clz.isAssignableFrom(acc.getClass())) {
```
je-ik: This looks expensive, and it will be called for each input element (if using counters). Can we move it into the `if (acc == null)` branch and perform the validation only once?

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/accumulators/FlinkNativeAccumulators.java
```diff
@@ -0,0 +1,175 @@
...
+    public void add(Duration duration) {
+      super.add(Math.toIntExact(duration.toMillis()));
+    }
+
+    // It is necessary to override clone method to instantiate this
+    // extended class and not the parent type.
+    @Override
+    @SuppressWarnings("CloneDoesntCallSuperClone")
+    public org.apache.flink.api.common.accumulators.Accumulator<Integer, TreeMap<Integer, Integer>> clone() {
+      FlinkTimer result = new FlinkTimer();
```
je-ik: Maybe this could be done by calling the constructor of the `FlinkTimer` with parameter `FlinkTimer`?

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/accumulators/FlinkNativeAccumulators.java
```diff
@@ -0,0 +1,175 @@
...
+        this.add(value);
+      }
+    }
+
+    // It is necessary to override clone method to instantiate this
+    // extended class and not the parent type.
+    @Override
+    @SuppressWarnings("CloneDoesntCallSuperClone")
+    public org.apache.flink.api.common.accumulators.Accumulator<Integer, TreeMap<Integer, Integer>> clone() {
+      FlinkHistogram result = new FlinkHistogram();
```
je-ik: Here also - it would be a little nicer to call a "copy constructor".

euphoria-flink/src/test/java/cz/seznam/euphoria/flink/testkit/FlinkExecutorProvider.java
```diff
@@ -32,6 +33,9 @@ default ExecutorEnvironment newExecutorEnvironment() throws Exception {
     RocksDBStateBackend backend = new RocksDBStateBackend("file://" + path);
     FlinkExecutor executor = new TestFlinkExecutor(ModuloInputSplitAssigner::new)
         .setStateBackend(backend);
+    // TODO should use some local test implementation of counters
+    // that is queryable in assertions
```
je-ik: Is this comment meant for the sake of unittesting user flows? Isn't this what `InMemExecutor` is for?

euphoria-operator-testkit/src/main/java/cz/seznam/euphoria/operator/test/MapElementsTest.java
je-ik: Why is this cast needed? What are the conflicting method declarations?

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/io/Context.java
vanekjar: Good question. It is the same. Basically the `Collector` should be directly implementing `Context`. It means collector just extends the context with a possibility to emit values.
The reason there is no direct relation between `Collector` and `Context` are lambda expressions that could become ambiguous and couldn't be used without casting.
```java
interface UnaryFunctor<IN, OUT> {
   apply(IN elem, Collector<OUT> collector);
}

UnaryFunctionEnv<IN, OUT> {
    OUT apply(IN what, Context context);
}
````

File: euphoria-flink/src/test/java/cz/seznam/euphoria/flink/testkit/FlinkExecutorProvider.java (hunk 12 above)
Comment (vanekjar) [inline euphoria-flink/src/test/java/cz/seznam/euphoria/flink/testkit/FlinkExecutorProvider.java]: It's rather intended for unit testing of the executor code. Just to make sure it provides accumulator adapter correctly in all UDFs. It is related to this issue #135

File: euphoria-flink/src/main/java/cz/seznam/euphoria/flink/accumulators/FlinkAccumulatorFactory.java (hunk 8 above)
Comment (vanekjar) [inline euphoria-flink/src/main/java/cz/seznam/euphoria/flink/accumulators/FlinkAccumulatorFactory.java]: I am not sure if I understand this correctly. What would be the return type of the factory then?

File: euphoria-operator-testkit/src/main/java/cz/seznam/euphoria/operator/test/MapElementsTest.java (hunk 6 above)
Comment (vanekjar) [inline euphoria-operator-testkit/src/main/java/cz/seznam/euphoria/operator/test/MapElementsTest.java]: I tried to dig deeper into Java to find out what are the candidates the compiler takes into account and the result is that I don't know. Kind of magic.
```
Error:(43, 13) java: reference to using is ambiguous
  both method <OUT>using(cz.seznam.euphoria.core.client.functional.UnaryFunction<IN,OUT>) in cz.seznam.euphoria.core.client.operator.MapElements.UsingBuilder and method <OUT>using(cz.seznam.euphoria.core.client.functional.UnaryFunctionEnv<IN,OUT>) in cz.seznam.euphoria.core.client.operator.MapElements.UsingBuilder match
```

File: euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/MapElements.java (hunk 5 above)
Comment (vanekjar) [inline euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/MapElements.java]: Good catch! The trick with `asContext` method works pretty well :+1:

File: euphoria-flink/src/main/java/cz/seznam/euphoria/flink/batch/BatchUnaryFunctorWrapper.java [hunk 13]
```diff
@@ -15,38 +15,51 @@
...
   public void flatMap(BatchElement<WID, IN> value,
-                      Collector<BatchElement<WID, OUT>> out)
-      throws Exception
-  {
-    f.apply(value.getElement(), new Context<OUT>() {
+                      org.apache.flink.util.Collector<BatchElement<WID, OUT>> out)
+          throws Exception {
+
+    AccumulatorProvider accumulators =
+            accumulatorFactory.create(settings, getRuntimeContext());
```
vanekjar: Good point :+1:

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/accumulators/FlinkNativeAccumulators.java
vanekjar: I am not sure about this. The clone method override is enforced by the Flink's superclass `Histogram`. Creating a copy constructor wouldn't solve anything in this case. But I agree this is a kind of ugly hack - at least properly described in code comment.

je-ik: I ment the three lines
```java
 FlinkTimer result = new FlinkTimer();
 result.merge(this);
 return result;
```
could read
```java
  return new FlinkTimer(this);
```

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/MapElements.java
xitep: cool! thaks for trying out. 👍

xitep:
thanks for the update. it looks good to me now 👍

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/io/Collector.java
```diff
@@ -29,4 +29,13 @@
...
   void collect(T elem);
+
+  /**
+   * Returns {@link Context} view of the collector.
+   * Since {@link Collector} usually share the same methods
+   * as {@link Context} it can be safely casted.
+   *
+   * @return this instance as a context class
+   */
+  Context asContext();
```
je-ik: This method seems that every implementation will have no other option than just to cast the `Collector` to `Context`, am I right? We could move that code inside either default implementation of this method, or maybe just cast it directly instead of calling this method.

vanekjar: It is not possible to directly cast `Collector` to `Context` since they are not subtypes of each other. The implementation of this doesn't rely on casting, see https://github.com/seznam/euphoria/pull/132/files#diff-e704b8f3c221db98ac7f7a4618484531R63

je-ik: Ah, right, my bad. :)

vanekjar:
Thank you both for the feedback. I resolved most of the mentioned issues with the newly added commit. From my point of view it's ready to be merged if you agree with the proposed changes.

Regarding the **performance** - I tested this branch with an added counter into the benchmark code and I couldn't find any difference regarding the running time compared to euphoria built from master branch.

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/io/Context.java
je-ik: This unfortunately implies that you will still have to specify the types of the lambda arguments, am I right? i.e.
```java
 FlatMap.of(...)
   .using((Integer in, Context ctx) -> {
   });
```
instead of just
```java
 FlatMap.of(...)
   .using((e, c) -> {
   });
```
Would it be possible to do something about that? I think that it would unfortunately mean that we would have to either add additional argument to the `UnaryFunctor`, or use other method name, e.g. `usingRich()`.
I'm aware that it would be more work. :-/
What do you think? And what do you think about naming the `Env` functions instead of `Env` suffix by `Rich` prefix?

xitep: Apart from the changes introduce by @vanekjar, there's troubles in general with `FlatMap` in regards to having to specify the types of the lambdas params. The generic parameter of the old "context" object (now it's called `Collector`) of the UDF is not bound by any preceding expression, and, thus, the user has to explicitly specify the types of the parameters explicitly.

If you check the usage of the `.using` method on `master` you'll see the pattern. That's why the change by Jara isn't that dramatic after all.

vanekjar: Could you please further develop the idea of `usingRich` method? What would it solve? Now the `using` is overloaded and I am fine with that. The mention `usingRich` would be just a replacement of overloading?

je-ik: Ah, I see your point. What I meant is that using overloading for methods of type `UnaryFunctor` and `UnaryFunctionEnv`, because the compiler cannot use type inference there. What is true is that you generally will not mix these two calls in single operator (the operator either uses `UnaryFunction` and `UnaryFunctionEnv` or `UnaryFunctor`. So it's fine. :+1:

xitep:
Thank you!

PR #137, opened 2017-05-30
#133 Accumulator API integrated to all executors

### LINKED ISSUE #133: Accumulators: Integrate Accumulator API with all available executors
Currently the Accumulator API is fully supported only in `euphoria-flink` executor. If any accumulator is accessed on other executors an `IllegalStateException` is raised. See `cz.seznam.euphoria.core.client.io.Environment`.

Goal of this issue to add support to all available executors and then it should be possible to remove default implementation with `FIXME` from the `Environment` class.

vanekjar:
I removed the `FIXME` from the `Environment` class and implemented the accumulator support to all collectors in both remaining executors (Spark, in-mem).

closes #133

### DIFF
--- a/euphoria-spark/src/main/java/cz/seznam/euphoria/spark/UnaryFunctorWrapper.java
+++ b/euphoria-spark/src/main/java/cz/seznam/euphoria/spark/UnaryFunctorWrapper.java
@@ -32,7 +32,7 @@ class UnaryFunctorWrapper<WID extends Window, IN, OUT>
   private final AccumulatorProvider.Factory accumulatorFactory;
   private final Settings settings;
 
-  private FunctionCollectorMem<OUT> cachedCollector;
+  private transient FunctionCollectorMem<OUT> cachedCollector;
 
   public UnaryFunctorWrapper(UnaryFunctor<IN, OUT> functor,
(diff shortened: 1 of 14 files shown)


euphoria-spark/src/main/java/cz/seznam/euphoria/spark/UnaryFunctorWrapper.java
```diff
@@ -26,31 +28,47 @@
 class UnaryFunctorWrapper<WID extends Window, IN, OUT>
         implements FlatMapFunction<SparkElement<WID, IN>, SparkElement<WID, OUT>> {
 
-  private final FunctionCollectorMem<OUT> context;
   private final UnaryFunctor<IN, OUT> functor;
+  private final AccumulatorProvider.Factory accumulatorFactory;
+  private final Settings settings;
 
-  public UnaryFunctorWrapper(UnaryFunctor<IN, OUT> functor) {
+  private FunctionCollectorMem<OUT> cachedCollector;
```
vanekjar: Yes, you're right. Theoretically this should be `transient`. It worked only because the collector is initially `null`. I'll fix it.

PR #139, opened 2017-06-02
#134 Implemented support for Spark native accumulators

### LINKED ISSUE #134: Accumulators: Implement support for Spark native accumulators
Goal of this ticket is to implement Spark custom `AccumulatorProvider` providing access to [Spark native accumulators](https://spark.apache.org/docs/latest/programming-guide.html#accumulators). 

Obviously this provider will be usable only for `euphoria-spark` executor. It is a similar situation to `FlinkNativeAccumulators` provider.

vanekjar:
closes #134

### DIFF
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/accumulators/VoidAccumulatorProvider.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/accumulators/VoidAccumulatorProvider.java
@@ -20,8 +20,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.time.Duration;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * Placeholder implementation of {@link AccumulatorProvider} that
@@ -49,7 +48,6 @@ public class VoidAccumulatorProvider implements AccumulatorProvider {
   }
 
   public static Factory getFactory() {
-    LOG.warn("Using accumulators with VoidAccumulatorProvider will have no effect");
     return Factory.get();
   }
 
@@ -62,6 +60,8 @@ public class VoidAccumulatorProvider implements AccumulatorProvider {
     private static final AccumulatorProvider PROVIDER =
             new VoidAccumulatorProvider();
 
+    private static final AtomicBoolean isLogged = new AtomicBoolean();
+
     private Factory() {}
 
     public static Factory get() {
@@ -70,6 +70,9 @@ public class VoidAccumulatorProvider implements AccumulatorProvider {
 
     @Override
     public AccumulatorProvider create(Settings settings) {
+      if (isLogged.compareAndSet(false, true)) {
+        LOG.warn("Using accumulators with VoidAccumulatorProvider will have no effect");
+      }
       return PROVIDER;
     }
   }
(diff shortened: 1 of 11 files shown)


euphoria-spark/src/main/java/cz/seznam/euphoria/spark/accumulators/SparkNativeAccumulators.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+  }
+
+  /**
+   * This class contains a map of accumulators. Spark requires all accumulators
+   * to be declared before the job is executed. This holder is created before
+   * the job is launched and registered to Spark context allowing Euphoria
+   * to add accumulators on the fly.
+   */
+  private static class SparkAccumulatorHolder
```
xitep: * looks good to me 👍 
* the only think i'm a bit worried about is the docs of `AccumulatorV2` stating ... _`OUT` should be a type that can be read atomically (e.g., Int, Long), or thread-safely (e.g., synchronized collections) because it will be read from other threads._ have you had any chance to investigate how accumulators are really accessed?

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/accumulators/SparkNativeAccumulators.java
```diff
@@ -0,0 +1,165 @@
...
+
+      return provider;
+    }
+
+    @Override
+    public void init(JavaSparkContext sparkContext) {
+      SparkAccumulatorHolder sparkAccumulatorHolder = new SparkAccumulatorHolder();
+      sparkContext.sc().register(sparkAccumulatorHolder, "accumulators");
+
+      provider = new SparkNativeAccumulators(sparkAccumulatorHolder);
```
xitep: please see my comment in the `SparkExecutor` regarding multi-flow execution

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/accumulators/VoidAccumulatorProvider.java
```diff
@@ -70,6 +67,7 @@ public static Factory get() {
 
     @Override
     public AccumulatorProvider create(Settings settings) {
+      LOG.warn("Using accumulators with VoidAccumulatorProvider will have no effect");
```
vanekjar: Agree :+1:

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/SparkExecutor.java
```diff
@@ -66,16 +67,31 @@ public void shutdown() {
...
+            Objects.requireNonNull(factory));
   }
 
   private Result execute(Flow flow) {
     if (!isBoundedInput(flow)) {
       throw new UnsupportedOperationException("Spark executor doesn't support unbounded input");
     }
 
+    // init accumulators
+    accumulatorFactory.init(sparkContext);
```
vanekjar: That is good point :+1: I haven't considered this use-case at first, my bad.
Seems the `accumulatorFactory` needs to be cloned for each flow and this shouldn't be a problem since the factory is serializable anyway. Please have a look at the newly added commit if it makes sense.

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/accumulators/SparkNativeAccumulators.java
vanekjar: Good catch! :+1:  I have actually missed that remark in Spark doc. I changed the internal structure to `ConcurrentHashMap`, that should solve the problem.

xitep:
Thanks!

PR #141, opened 2017-06-15
#! Fix a few Javadoc and Findbugs warnings

vanekjar:


benchmarks/native-beam/native-beam-spark/src/main/java/cz/seznam/euphoria/benchmarks/beam/StatelessJavaSerializer.java
```diff
@@ -64,8 +63,8 @@ public void write(Kryo kryo, Output output, Object object) {
 
   @SuppressWarnings("unchecked")
   public Object read (Kryo kryo, Input input, Class type) {
-    try {
-      return new ObjectInputStreamWithClassLoader(input, kryo.getClassLoader()).readObject();
+    try (ObjectInputStream objectStream = new ObjectInputStreamWithClassLoader(input, kryo.getClassLoader())) {
+      return objectStream.readObject();
     } catch (Exception e) {
       throw new KryoException("Error during Java deserialization.", e);
     }
```
xitep: i'm not sure these two changes are correct. newly, the change will cause the `output/input` to be closed. and i believe this will prevent writing/reading two or more values to/from the output/input.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/batch/SortTranslator.java
```diff
@@ -90,7 +90,7 @@ public DataSet translate(
     return sorted;
   }
   
-  public class SortPartitionerWrapper<T> 
+  public static class SortPartitionerWrapper<T>
```
xitep: good catch! 👍

benchmarks/native-beam/native-beam-spark/src/main/java/cz/seznam/euphoria/benchmarks/beam/StatelessJavaSerializer.java
vanekjar: Many thanks, you're right, it would cause the Kryo `input/output` to be closed too early. I reverted this change and added explaining comment.

xitep: ok, many thanks for the confirmation! 👍 i'll merge as soon as travis gives us green light.

PR #144, opened 2017-07-30
[WIP] Beam runtime

dmvk:


euphoria-beam/src/main/java/cz/seznam/euphoria/beam/io/BeamWriteSink.java
```diff
@@ -0,0 +1,99 @@
...
+      this.sink = sink;
+    }
+
+    @Setup
+    public void setup() {
+      writer = sink.openWriter(partitionId);
+    }
+
+    @StartBundle
+    public void startBundle(StartBundleContext c) throws IOException {
```
je-ik: You can just drop this method, if it is supposed to ignore the start bundle info.

PR #151, opened 2017-08-07
Spark broadcast join

dmvk:


euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Join.java
```diff
@@ -129,7 +129,10 @@
     private final UnaryFunction<LEFT, KEY> leftKeyExtractor;
     private final UnaryFunction<RIGHT, KEY> rightKeyExtractor;
     private final BinaryFunctor<LEFT, RIGHT, OUT> joinFunc;
-    private boolean outer;
+
+    private boolean outer = false;
+    private boolean leftCached = false;
+    private boolean rightCached = false;
```
je-ik: I would drop this, because it just brings unnecessary implementation complexity and copy&pasting.

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Join.java
```diff
@@ -234,7 +255,10 @@ public static OfBuilder named(String name) {
   private final BinaryFunctor<LEFT, RIGHT, OUT> functor;
   final UnaryFunction<LEFT, KEY> leftKeyExtractor;
   final UnaryFunction<RIGHT, KEY> rightKeyExtractor;
-  boolean outer = false;
+
+  boolean outer;
+  boolean leftCached;
+  boolean rightCached;
```
je-ik: All `final`?

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/FlatMapFunctionWithContext.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  FunctionCollectorMem<OUT> getContext() {
+    if (cachedCollector == null) {
+      cachedCollector = new FunctionCollectorMem<>(accumulators);
+    }
+    return cachedCollector;
+  }
+
+  public interface InnerFunction<IN, OUT> extends Serializable {
```
je-ik: Leaking public interface from non public class.

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/FlatMapFunctionWithContext.java
```diff
@@ -0,0 +1,53 @@
...
+    this.function = Objects.requireNonNull(function);
+    this.accumulators = Objects.requireNonNull(accumulators);
+  }
+
+  @Override
+  public Iterator<OUT> call(IN in) throws Exception {
+    return function.call(in, getContext());
+  }
+
+  FunctionCollectorMem<OUT> getContext() {
```
je-ik: This looks strange, why `getContext()` returns `FunctionCollectorMem`? :)

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/JoinTranslator.java
```diff
@@ -0,0 +1,128 @@
...
+      // ~ pass partitioner implementation
+      if (operator.isOuter()) {
+        joined = leftPair.fullOuterJoin(rightPair, new PartitioningWrapper(partitioning))
+            .mapToPair(t -> new Tuple2<>(t._1, new Tuple2<>(t._2._1.orNull(), t._2._2.orNull())));
+      } else {
+        joined = leftPair.join(rightPair, new PartitioningWrapper(partitioning));
+      }
+    }
+
+    // @TODO ?????? I DON'T KNOW WHAT I'M DOING
```
je-ik: :)

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/JoinTranslator.java
```diff
@@ -0,0 +1,128 @@
...
+    final LazyAccumulatorProvider acc =
+        new LazyAccumulatorProvider(context.getAccumulatorFactory(), context.getSettings());
```
je-ik: We have to look deeper into this. I don't like the accumulator abstraction.

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/JoinTranslator.java
```diff
@@ -0,0 +1,128 @@
...
+          t._2._1 == null ? null : t._2._1.getElement(),
+          t._2._2 == null ? null : t._2._2.getElement(),
+          collector);
+      return Iterators.transform(collector.getOutputIterator(), e ->
+          new SparkElement(window, maxTimestamp, Pair.of(t._1.key(), e)));
+    }, acc));
+  }
+
+  /**
+   * Expand element to all windows
```
je-ik: What is this comment supposed to mean? :)

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/JoinTranslator.java
```diff
@@ -0,0 +1,128 @@
...
+      // ~ shit just got real ... @todo discuss operator test kit
+      final Iterable<Window> windows = windowing.assignWindowsToElement(new SparkElement(se.getWindow(),
+          se.getTimestamp(), left ? Either.left(se.getElement()) : Either.right(se.getElement())));
+      final List<Tuple2<KeyedWindow, SparkElement>> out = new ArrayList<>();
+      for (Window window : windows) {
+        out.add(new Tuple2<>(
+            new KeyedWindow<>(window, se.getTimestamp(), keyExtractor.apply(se.getElement())),
+            new SparkElement(window, se.getTimestamp(), se.getElement())));
+      }
+      return out.iterator();
```
je-ik: Here is unnecessary placing elements into new collection. For performance reasons it would be better to use something like `Iterators.transform`.

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Join.java
```diff
@@ -129,7 +129,9 @@
     private final UnaryFunction<LEFT, KEY> leftKeyExtractor;
     private final UnaryFunction<RIGHT, KEY> rightKeyExtractor;
     private final BinaryFunctor<LEFT, RIGHT, OUT> joinFunc;
-    private boolean outer;
+
+    private Type type = Type.INNER;
```
je-ik: :+1:

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Join.java
```diff
@@ -460,7 +486,15 @@ public void mergeFrom(EarlyEmittingJoinState other) {
...
-    return outer;
+    return type == Type.OUTER;
+  }
+
+  public Type getType() {
+    return type;
+  }
+
+  public boolean isLeftCached() {
+    return leftCached;
```
je-ik: This looks is should read `type == LEFT && leftCached`

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/BroadcastJoinTranslator.java
```diff
@@ -0,0 +1,176 @@
...
+import scala.Tuple2;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+public class BroadcastJoinTranslator implements SparkOperatorTranslator<Join> {
```
je-ik: Maybe some minor comment about purpose of this class?

je-ik: The documentation in overall will deserve attention, I think we should keep as much documentation as possible in class describing comments, because that way we can better keep it up-to-date.

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/SparkFlowTranslator.java
```diff
@@ -51,114 +52,101 @@
...
+    Translation.add(translations, ReduceByKey.class, new ReduceByKeyTranslator(),
+        ReduceByKeyTranslator::wantTranslate);
 
-    // derived operators
-    Translation.set(translations, ReduceByKey.class, new ReduceByKeyTranslator(),
-            ReduceByKeyTranslator::wantTranslate);
+    // ~ batch broadcast join for a very small left side
+    Translation.add(translations, Join.class, new BroadcastJoinTranslator(), o ->
+        o.getHints().contains(JoinHints.broadcastHashJoin())
+            && (o.getWindowing() == null || o.getWindowing() instanceof GlobalWindowing));
```
je-ik: This look odd. Why do we constraint usability of this hint to to global or attached windowing? There should be no reason for that.

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/SparkFlowTranslator.java
```diff
@@ -51,114 +52,101 @@
...
-      Preconditions.checkState(
-              tx.accept == null || Boolean.TRUE.equals(tx.accept.apply(op)));
-
-      JavaRDD<?> out = tx.translator.translate(op, executorContext);
-
-      // save output of current operator to context
-      executorContext.setOutput(op, out);
+      Translation firstMatch = null;
+      for (Translation tx : txs) {
+        if (tx.accept == null || Boolean.TRUE.equals(tx.accept.apply(op))) {
```
je-ik: Why not just `|| tx.accept.apply(op)`? Although I'm aware that this code wasn't written in this PR.

euphoria-spark/src/test/java/cz/seznam/euphoria/spark/BroadcastHashJoinTest.java
```diff
@@ -0,0 +1,104 @@
...
+import cz.seznam.euphoria.spark.testkit.SparkExecutorProvider;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+
+@RunWith(ExecutorProviderRunner.class)
+public class BroadcastHashJoinTest extends AbstractOperatorTest implements SparkExecutorProvider {
```
je-ik: Strongly suggest moving this class to operator test kit and run against all executors. It *must* pass on all of them, regardless if they support, or don't support the hint. It would be good to actually run the same test twice - with hint and without hint and expect the same results.

je-ik: Although it looks like basically the same test is already there. Maybe we can just add another one with the hint. Maybe we could parametrize the join test somehow. I'm not sure how, although. :)

PR #153, opened 2017-08-20
#152 [euphoria-core] remove partitioning from stateful operators

je-ik:


je-ik:
Replaced via work in branch `160-drop-explicit-partitioning` and issue #160.

PR #165, opened 2017-11-01
 [euphoria] reformat licenses

je-ik:


je-ik:
Oops, this is definitely what was intended. :)

PR #176, opened 2017-11-09
 [euphoria-core] add API audience annotations to core

je-ik:


je-ik:
Changed to `Audience`, the build failed due to missing header, it is fixed now and should proceed OK.

PR #178, opened 2017-11-09
Move to gradle init

dmvk:


.travis.yml
```diff
@@ -1,29 +1,28 @@
 sudo: false
-
 language: java
 
 jdk:
-- oraclejdk8
-
-addons:
```
je-ik: I'd keep this here, because it installs newest JDK available (if I understand it correctly).

.travis.yml
```diff
@@ -1,29 +1,28 @@
...
-  - secure: MB8I+BHVP9QMaJQF1ALbp+/n0rfU1EVaPPOthn4MRoy6AUTL5Aigh/qJ5zghBR1yOOkbREJhgNkaN9Jwk8Oncxpl34FYj2+Sx9YPh2SB1q04k2VBeoP4lcn26ZMyGQJRjGSm37jwF/N2pEya7auYOo5EGwVFCGPndnA4YngsGiOfnlYWboDNp37RhqXcME6epo0pd/QYuJSGkjd5pMdEpenWdVmXkAxRLxFhgeKi7oI8sMZ8pR5I/i5jDlPHGlk2t9u9wWDkz7rYwP3caO6s3lnlRBCgvDc04Tvh8X+s6jOoaBK3MNeNkujvUOSkJT8z9VHHYc/BgpOJ/xhqAnqrKaJgwzgAf3PcXT1VNmL2tW/dT/tJXJ7V/w3wgUUIl3+rDZMJ1qePlkrkyvt8IKSFsf/V0Dplj7SHpr7RMPC6k3cr0rUzPndLK0UIF7aSJ7qBMFHn+QVNTlppSxZjkq9OT690jbguo49Y6Clm50JPw0sLG9TTqvfWA36DuHZPrxUXHnqRwkkUe1aM6bQuZcHzfE3/fDcxh63RNzraN49akAT7x6Al8vrdgGJmDnp0gqoD5lIZtQtbNkBfOLkcD0ZEIgQaxZ4BVn9xmrg3JxVm5OfofSXYsBh6dfXQ8MYThZ/dWmBxnnTwXmJhtwoCwmYDMgMNOWfZeIy/dI346RkqGpk=
+#cache:
+#  directories:
+#  - ~/.m2/repository
+#
+#after_success:
+#- ./cd/before-deploy.sh
+#- ./cd/deploy.sh
```
je-ik: I think we should not comment things out, let's just delete it and use git for retrieving it, if needed in the future.

benchmarks/benchmark-beam-flink/build.gradle
```diff
@@ -0,0 +1,6 @@
+dependencies {
+  compile project(':benchmarks:benchmark-beam-common')
+  compile "org.apache.flink:flink-java:${beamFlinkVersion}"
+  compile "org.apache.flink:flink-clients_${beamScalaVersion}:${beamFlinkVersion}"
+  compile "org.apache.flink:flink-streaming-java_${beamScalaVersion}:${beamFlinkVersion}"
```
je-ik: No findbugs in here?

benchmarks/benchmark-beam-spark/build.gradle
```diff
@@ -0,0 +1,6 @@
+dependencies {
+  compile project(':benchmarks:benchmark-beam-common')
+  compile "org.apache.beam:beam-runners-spark:${beamVersion}"
+  compile "org.apache.spark:spark-core_${beamScalaVersion}:${beamSparkVersion}"
+  compile "org.apache.spark:spark-streaming_${beamScalaVersion}:${beamSparkVersion}"
```
je-ik: Findbugs?

benchmarks/benchmark-euphoria-common/build.gradle
```diff
@@ -0,0 +1,7 @@
+dependencies {
+  compile project(':benchmarks:benchmark-model')
+  compile project(':euphoria-core')
+  compile project(':euphoria-hadoop')
+  compile project(':euphoria-kafka')
+  compile "com.typesafe:config:${configVersion}"
```
je-ik: Maybe findbugs should be in all dependencies ... Or could it be made inherited from some sort of parent?

build.gradle
```diff
@@ -0,0 +1,93 @@
...
+
+subprojects {
+
+  apply plugin: 'java'
+  apply plugin: 'com.github.hierynomus.license'
+  apply plugin: 'com.github.spotbugs'
+  apply from: "${rootProject.projectDir}/gradle/versions.gradle"
+
+  license {
+    include '**/*.java'
```
je-ik: Not sure about this ... first of all, the maven license plugin checked much more files, including shell scripts, XML files, and many more ... we shouldn't drop this functionality.
Second doubt is about the year, I think it should stay as it is - i.e. `2016-<currentYear>`.

build.gradle
```diff
@@ -0,0 +1,93 @@
...
+    ext.year = Calendar.getInstance().get(Calendar.YEAR)
+    ext.name = 'Seznam.cz, a.s.'
+  }
+
+  repositories {
+    mavenCentral()
+  }
+
+  dependencies {
+    compileOnly "org.projectlombok:lombok:${lombokVersion}"
```
je-ik: Lombok should definitely not be top level dependency. Euphoria itself deliberately chose not to use it at all (as I see, it is as a dependency in `euphoria-hadoop`, it should be removed). I see it also in root pom, but it is not used in any code.

build.gradle
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  test {
+    testLogging {
+      events 'skipped', 'failed'
+    }
+  }
+
+  tasks.withType(SpotBugsTask) {
+    // ~ TODO: do not ignore failures and decrease report level
```
je-ik: This comment look serious and we should fix it before merging.

build.gradle
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  gradle.projectsEvaluated {
+    tasks.withType(JavaCompile) {
+      options.compilerArgs << "-Xlint:unchecked" << "-Xlint:deprecation"
+    }
+  }
+}
+
+// ~ configure euphoria sub-projects (no benchmarks)
```
je-ik: type `no benchmarks` -> `excluding benchmarks`

build.gradle
```diff
@@ -0,0 +1,93 @@
...
+
+configure(subprojects.findAll { it.name.startsWith('euphoria-') }) {
```
je-ik: If it is excluding benchmarks, what about `findAll { it.name != "benchmarks" }`? :)

euphoria-core/src/main/java/cz/seznam/euphoria/core/annotation/operator/Basic.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2016-2017 Seznam.cz, a.s.
+ * Copyright 2017 Seznam.cz, a.s.
```
je-ik: :-1: The year should remain as was previously.

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/dataset/windowing/Session.java
```diff
@@ -21,7 +21,7 @@
 import cz.seznam.euphoria.core.client.triggers.TimeTrigger;
 import cz.seznam.euphoria.core.client.triggers.Trigger;
 import cz.seznam.euphoria.core.client.util.Pair;
-import cz.seznam.euphoria.shaded.guava.com.google.common.base.Preconditions;
+import com.google.common.base.Preconditions;
```
je-ik: :+1: for this ... I never liked the shaded guava as direct dependency.

euphoria-hadoop/build.gradle
```diff
@@ -0,0 +1,7 @@
+dependencies {
+  compile project(':euphoria-core')
+  compile "org.apache.hadoop:hadoop-common:${hadoopVersion}"
+  compile "org.apache.hadoop:hadoop-mapreduce-client-core:${hadoopVersion}"
+  compile "com.esotericsoftware.kryo:kryo:${kryoVersion}"
+//  testCompile "org.apache.hadoop:hadoop-client:${hadoopVersion}"
```
je-ik: Do we need it or not? :)

benchmarks/benchmark-beam-flink/build.gradle
dmvk: included from benchmark-beam-common

benchmarks/benchmark-beam-spark/build.gradle
dmvk: included from benchmark-beam-common

benchmarks/benchmark-euphoria-common/build.gradle
dmvk: findbugs are applied to all subprojects in root build.gradle (dependency is need only in case you need to add suppressing annotations)

build.gradle
dmvk: it should be part of another pull request, so the reported issues will become more visible.

build.gradle
dmvk: ok, ill remove it for now, but we should reconsider ;)

.travis.yml
dmvk: removed

build.gradle
dmvk: done

build.gradle
dmvk: done

Issue #157, opened 2017-10-18
Outputs have to have option to modify the Flow


je-ik:
Some outputs need a specific transformation to be applied directly before the output is written to the sink. Example of this is HFileOutput, which needs to do a groupBy operation on region ID, sort values by KeyValueComparator and then flush it to the sink.

We need to come up with a consistent and clean way to provide outputs with the possiblity.

je-ik:
Closed via #180

PR #180, opened 2017-11-10
157 enable outputs to modify flow

je-ik:


je-ik:
Closes #157

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/io/DataSink.java
```diff
@@ -57,4 +58,19 @@ default void initialize() {}
...
+
+  /**
+   * Called when adding output sink to the resulting DAG for processing.
+   * Purpose of this method is to enable sink to add arbitrary transformations
+   * to the dataset, before it gets persisted via this sink.
+   *
+   * @param output the dataset being written by this sink
+   * @return true if the flow was modified
+   */
+  default boolean onAdded(Dataset<T> output) {
```
dmvk: I guess we should find a better name for this :( Something like `prepareDataset`, `modifyFlow`...

je-ik:
Fixed. Please mark as ready to merge.

PR #186, opened 2017-11-14
Do not ignore spotbugs failures.

dmvk:
We should increase report level in the near future.

### DIFF
--- a/euphoria-local/src/main/java/cz/seznam/euphoria/executor/local/LocalExecutor.java
+++ b/euphoria-local/src/main/java/cz/seznam/euphoria/executor/local/LocalExecutor.java
@@ -15,6 +15,7 @@
  */
 package cz.seznam.euphoria.executor.local;
 
+import com.google.common.collect.Iterables;
 import cz.seznam.euphoria.core.client.accumulators.AccumulatorProvider;
 import cz.seznam.euphoria.core.client.accumulators.VoidAccumulatorProvider;
 import cz.seznam.euphoria.core.client.dataset.windowing.GlobalWindowing;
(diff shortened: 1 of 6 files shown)


euphoria-local/src/main/java/cz/seznam/euphoria/executor/local/LocalExecutor.java
```diff
@@ -719,17 +717,19 @@ private InputProvider execReduceStateByKey(
...
                   targetWindows = Collections.singleton(datum.getWindow());
                 }
 
                 if (!isMergingWindowing && Iterables.size(targetWindows) == 1) {
-                  windowShift = new Random(
-                      Iterables.getOnlyElement(targetWindows).hashCode()).nextInt();
+                  final int windowHash = Iterables.getOnlyElement(targetWindows).hashCode();
+                  synchronized (random) {
+                    windowShift = shifts.computeIfAbsent(windowHash, (ignore) -> random.nextInt());
+                  }
```
je-ik: I'm not sure, that we are trading time for space at reasonable exchange ratio. :)
The original code might be a little more CPU ang GC intensive, while this code will have to hold possible unbounded map in memory. I vote for the previous implementation.

euphoria-local/src/main/java/cz/seznam/euphoria/executor/local/WatermarkEmitStrategy.java
```diff
@@ -39,11 +42,6 @@ public void schedule(Runnable action) {
       scheduler.scheduleAtFixedRate(action, 100, 100, TimeUnit.MILLISECONDS);
     }
 
-    @Override
-    public void close() {
-      scheduler.shutdown();
-    }
-
```
je-ik: Why we don't need to shutdown the executor anymore?

dmvk: it is a static deamonthread executor, so it will shut down automatically when jvm stops. And nobody was using this method anyway :)

euphoria-local/src/main/java/cz/seznam/euphoria/executor/local/LocalExecutor.java
dmvk: How about setting an upper bound for this? `windowShift = shifts.computeIfAbsent(windowHash % BOUND, (ignore) -> random.nextInt());`

PR #195, opened 2017-11-25
 [euphoria-core] bundle guava into core and shade it

je-ik:


je-ik:
I think this kind of works, but the module `euphoria-core` still depends on guava, which is wrong. I didn't figure out, how to safely remove it, though. Any ideas welcome!

Issue #181, opened 2017-11-10
Add HBase module


je-ik:
Add `HBaseDataSource`, `HBaseDataSink`, `HFileDataSink`.

je-ik:
Closed via #205

PR #205, opened 2017-12-01
 [euphoria-hbase] #181 add bulk loading to hbase sink

je-ik:


euphoria-hbase/src/main/java/cz/seznam/euphoria/hbase/HFilesSink.java
```diff
@@ -123,10 +131,19 @@ public HFilesSink build() {
 
   }
 
+  private final String table;
   private final byte[][] endKeys;
 
+  @VisibleForTesting
+  HFilesSink(HFilesSink clone) {
```
dmvk: should the class name be HFileSink?

dmvk:
LGMT. Good job!

euphoria-hbase/src/main/java/cz/seznam/euphoria/hbase/HFilesSink.java
```diff
@@ -192,22 +212,72 @@ public boolean prepareDataset(
           }
         })
         .withSortedValues(comparator::compare)
+        .output();
+
+    String outputDir = HFilesSink.this.getConfiguration().get(FileOutputFormat.OUTDIR);
+
+    ReduceWindow.of(hfiles)
```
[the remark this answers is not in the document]
je-ik: Hm, making it optional would change a lot the underlaying logic. What should happen when the bulk load doesn't occur? The `ReduceWindow` is at the moment necessary, because `DataSink.commit()` is never called during the execution, so the resulting HFiles otherwise remain somewhere in `_temporary` directory. What I can do is make configurable the call to `loadIncrementalHFiles`, so that the hfiles would simply accumulate in the target directory. But the reducing part is necessary.

euphoria-hbase/src/main/java/cz/seznam/euphoria/hbase/HFilesSink.java
je-ik: so be it

je-ik:
Closes #181

je-ik:
Sorry, I accidentally ammended the review commit to the main commit. My bad. :-(

gradle/versions.gradle
```diff
@@ -4,7 +4,7 @@ ext {
   guavaVersion = '23.0'
   hadoopVersion = '2.7.4'
   hbaseVersion = '1.3.1'
-  kafkaVersion = '0.10.0.1'
+  kafkaVersion = '0.11.0.0'
```
dmvk: We should be really careful about this, as kafka module does not have any test cases ... (we should completely rewrite it in order to take full advantage of new kafka-client apis)

euphoria-hbase/src/main/java/cz/seznam/euphoria/hbase/HFileSink.java
```diff
@@ -192,22 +229,80 @@ public boolean prepareDataset(
...
+
+    String outputDir = HFileSink.this.getConfiguration().get(FileOutputFormat.OUTDIR);
+
+    ReduceWindow.of(hfiles)
+        .valueBy(Pair::getSecond)
+        .reduceBy(s -> {
+          Path o = new Path(outputDir);
+          s.forEach(p -> moveTo(p, o));
+          loadIncrementalHFiles(o);
+          return "";
```
dmvk: this seems like ReduceWindow api needs some attention :)

Issue #143, opened 2017-06-29
Complete join semantics


je-ik:
In contrast to purely batch join, in stream processing the variety of join types is much broader. The main differences are:
 * streams have time ordering (as opposed to batch, where all elements have the same time)
 * streams can be windowed (as opposed to batch, where all elements belong to the same window)

This implies more options for joins besides the "classical" - left, right, inner, full outer:
 * join can either apply windowing on both streams or can cache left or right stream locally (`windowed`, `leftCached`, `rightCached`)
 * because of the time ordering the join can be performed in fashion of cartesian product of left and right elements having the same key, or only defined element from each set, yielding `leftFirst`, `leftLast`, `leftAll`, `rightFirst`, `rightLast`, `rightAll`

Of course, some combinations don't make sense, so we have to make the settings applicable only on meaningful combinations, which are:
 * {left, right, inner, full}, windowed, {left, right}{First, Last, All}
 * left, rightCached, right{First, Last}
 * right, leftCached, left{First, Last}
 * {inner, fullOuter}, leftCached, left{First, Last}
 * {inner, fullOuter}, rightCached, right{First, Last}

The client code could look like this for classical windowed left join taking last element in each window
```java
  Join.of(left, right)
    .using(...)
    .left()
    .windowed(/* windowing comes here */)
    .leftLast()
    .rightLast()
    .output();
```
The following code will function as equivallent to joining `KStream` with `KTable` in Kafka streams:
```java
  Join.of(left, right)
    .using(...)
    .left()
    .rightCached()
    .rightLast()
    .output()
```

PR #206, opened 2017-12-01
Join operator API improvements.

dmvk:


dmvk:
Related to #143

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/InnerJoin.java
```diff
@@ -0,0 +1,99 @@
...
+
+import cz.seznam.euphoria.core.annotation.audience.Audience;
+import cz.seznam.euphoria.core.client.dataset.Dataset;
+import cz.seznam.euphoria.core.client.functional.BinaryFunctor;
+import cz.seznam.euphoria.core.client.functional.UnaryFunction;
+
+import java.util.Objects;
+
+@Audience(Audience.Type.CLIENT)
+public class InnerJoin {
```
je-ik: This operator should remain just `Join`.

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/RightJoin.java
```diff
@@ -0,0 +1,102 @@
...
+import cz.seznam.euphoria.core.annotation.audience.Audience;
+import cz.seznam.euphoria.core.client.dataset.Dataset;
+import cz.seznam.euphoria.core.client.functional.BinaryFunctor;
+import cz.seznam.euphoria.core.client.functional.UnaryFunction;
+
+import java.util.Objects;
+import java.util.Optional;
+
+@Audience(Audience.Type.CLIENT)
+public class RightJoin {
```
je-ik: How is this implementation different from `LeftJoin`? Isn't this just quite expensive syntactic sugar around the same implementation with swapped arguments?

euphoria-operator-testkit/src/main/java/cz/seznam/euphoria/operator/test/JoinTest.java
```diff
@@ -101,6 +143,45 @@ public void batchJoinOuter() {
...
         return Arrays.asList(
             Pair.of(0, "0+null"),
+            Pair.of(2, "2+12"), Pair.of(2, "2+12"), Pair.of(4, "4+14"),
+            Pair.of(1, "1+11"), Pair.of(1, "1+11"),
+            Pair.of(3, "3+13"), Pair.of(3, "3+13"));
+      }
+    });
+  }
+
+  @Processing(Processing.Type.BOUNDED)
```
je-ik: Why do we run these tests only for bounded datasets?

je-ik: OK, I see, the reason is probably that unbounded dataset cannot be join using global windowing. There are separate tests for windowing.

PR #207, opened 2017-12-02
[euphoria-core] Union can accept more than two data sets

dmvk:


euphoria-flink/src/main/java/cz/seznam/euphoria/flink/batch/UnionTranslator.java
```diff
@@ -31,13 +31,13 @@
   public DataSet translate(
       FlinkOperator<Union> operator,
       BatchExecutorContext context) {
-
-    List<DataSet> inputs = (List) context.getInputStreams(operator);
-    if (inputs.size() != 2) {
+    final List<DataSet> inputs = (List) context.getInputStreams(operator);
+    if (inputs.size() < 2) {
```
je-ik: Is this really an error? I think that the operator could gracefully reduce to no-op with single input.

dmvk: I feel that this implies that business logic is somehow wrong and should fail right away.

PR #208, opened 2017-12-02
Update readme.

dmvk:


je-ik:
I think we should definitely have some other way how to be contacted besides issues. But I guess we can postpone this now.

PR #209, opened 2017-12-02
[euphoria-operator-testkit] Reduce by key bounded parallel test.

dmvk:


euphoria-operator-testkit/src/main/java/cz/seznam/euphoria/operator/test/ReduceByKeyTest.java
```diff
@@ -337,13 +339,52 @@ public void testReduceWithoutWindowing() {
     });
   }
 
+  @Processing(Processing.Type.BOUNDED)
```
je-ik: This looks odd, why not reduce the sorted values to list and check the final list?

PR #211, opened 2017-12-07
[euphoria-core] make sure that windowBy() can be appliedIf()

je-ik:
Purpose of this PR is to make sure, that every call to `windowBy` can be wrapped into `applyIf()` call. This is very handy for cases when you have optional windowing strategy for a RSBK type operation, e.g.
```java
  @Nullable Windowing windowing = ...;
  ReduceByKey.of(...)
    .keyBy(..)
    ...
    .applyIf(windowing != null, b -> b.windowBy(windowing))
    .output();
```

The alternative approach would be either:
 * enable passing `null` to `windowBy` and handle it there, which seems unclean
 * let it to the user to work with the builders, storing them to intermediate variables and branching the code himself, which is just ugly

This PR implies that the result build of `applyIf` and `windowBy` is the same, which has some implications that are not very nice - i.e. you can apply `windowBy` multiple times, but that is tolerable imo.

### DIFF
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Distinct.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Distinct.java
@@ -44,8 +44,7 @@ import java.util.Objects;
 )
 public class Distinct<IN, ELEM, W extends Window>
     extends StateAwareWindowWiseSingleInputOperator<
-        IN, IN, IN, ELEM, ELEM, W, Distinct<IN, ELEM, W>>
-{
+        IN, IN, IN, ELEM, ELEM, W, Distinct<IN, ELEM, W>> {
 
   public static class OfBuilder implements Builders.Of {
     private final String name;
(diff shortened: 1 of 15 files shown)


euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Distinct.java
```diff
@@ -61,14 +61,14 @@
...
-          implements Builders.WindowBy<IN>,
-                     Builders.Output<ELEM>,
-                     OptionalMethodBuilder<WindowingBuilder<IN, ELEM>>
-  {
-    private final String name;
-    private final Dataset<IN> input;
+      implements Builders.WindowBy<IN, WindowingBuilder<IN, ELEM>>,
+      Builders.Output<ELEM>,
+      OptionalMethodBuilder<WindowingBuilder<IN, ELEM>> {
+
```
dmvk: whitspace

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/TopPerKey.java
```diff
@@ -178,20 +178,22 @@ public void mergeFrom(MaxScored<V, C> other) {
...
-    private String name;
-    private final Dataset<IN> input;
-    private final UnaryFunction<IN, K> keyFn;
-    private final UnaryFunction<IN, V> valueFn;
-    private final UnaryFunction<IN, S> scoreFn;
+    final String name;
+    final Dataset<IN> input;
+    final UnaryFunction<IN, K> keyFn;
+    final UnaryFunction<IN, V> valueFn;
+   final UnaryFunction<IN, S> scoreFn;
```
dmvk: reformat

euphoria-core/src/test/java/cz/seznam/euphoria/core/client/operator/DistinctTest.java
```diff
@@ -72,4 +72,15 @@ public void testBuild_Windowing() {
     assertTrue(distinct.getWindowing() instanceof Time);
   }
 
+  // this doesn't need @Test, it just tests compilability
```
dmvk: i'd say it should be regular test

PR #213, opened 2017-12-10
WIP: Side outputs support

dmvk:


dmvk:
This should be part of the Beam DSL implementation.

Issue #131, opened 2017-05-17
euphoria-core: Add access to timestamp and key to context


je-ik:
* add option to access element's timestamp from `Context` via `Context#getTimestamp()`
* extend `Context` to `KeyedContext` with `KeyedContext#getKey` and pass it to keyed operators (RSBK, RBK, join)

PR #214, opened 2017-12-11
Added PrefetchIterator

t-novak:
Useful for construct like:
```java
Dataset<T> data = Join.of(input, config)...;
ReduceByKey.of(data)
  .reduceBy((Stream<T> values, Collector<X> collector) -> {
    final PrefetchIterator<T> iterator = new PrefetchIterator<>(values);
    T first = iterator.prefetch();
    if (first != null) {
      if (first.selected())
        processSelected(iterator, collector);
      else
        processUnselected(iterator, collector);
    }
  });
```

### DIFF
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/util/PrefetchIterator.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/util/PrefetchIterator.java
@@ -1,0 +1,65 @@
+/**
+ * Copyright 2016-2017 Seznam.cz, a.s.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package cz.seznam.euphoria.core.util;
+
+import java.util.Iterator;
+import java.util.stream.Stream;
+
+/**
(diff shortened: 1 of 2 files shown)


euphoria-core/src/main/java/cz/seznam/euphoria/core/util/PrefetchIterator.java
```diff
@@ -0,0 +1,65 @@
...
+      return it.next();
+    }
+  }
+
+  /**
+   * Get next value without moving the iterator forward.
+   *
+   * @return Next value (or null on the end)
+   */
+  public T prefetch() {
```
je-ik: Can be name this `peek`? That is compliant with naming conventions of other methods performing this task.

t-novak: I'm in doubt about `peek()`.

It has different meaning in java 8 stream API and maybe someone would like to add `PeekingStream` later doing the same thing on stream.

je-ik: I'd say, that the naming of `Stream#peek` seems unfortunate, but OK. Can we come up with some other naming that would be clear to users? I have a strong feeling that the current naming is not self-explanatory.

t-novak: OK, more names:

- `LookupIterator::lookup()`
- `PeekingIterator::peek()`
- `PeepIterator::peep()`
- `SurveyIterator::survey()`
- `InspectIterator::inspect()`
- ?

je-ik: I still vote for `peek`, but maybe others might have some opinion on this.

je-ik: @dmvk Cool, I wasn't aware of that. I think `Iterators.peekingIterator(values.iterator())` does exactly this and we shouldn't reinvent the wheel. Also, as discussed over slack, motivation for this is probably missing #131 
We should probably close this.

PR #216, opened 2017-12-11
[euphoria-operator-testkit] Filter test.

dmvk:


dmvk:
What should be expected behaviour, in case persist was not call? I'd say that its subtree should not be part of the result DAG.

je-ik:
The expected behavior is quite undefined I suppose. It definitely should not fail. :)
I'm not sure it the operator should not be run at all, because it could have some side-effects. On the other hand, if it has side-effects, it can be persisted into `VoidSink` for instance. But it might seem a little weird to users.

je-ik:
Do we still want to merge this?

dmvk:
obsolete

PR #218, opened 2017-12-12
Added outputValues() to operators producing pairs

t-novak:


euphoria-core/src/test/java/cz/seznam/euphoria/core/client/operator/JoinTest.java
```diff
@@ -57,6 +57,32 @@ public void testBuild() {
...
+
+    assertEquals(flow, joined.getFlow());
+    assertEquals(2, flow.size());
+
+    Join join = (Join) flow.operators().iterator().next();
+    assertEquals(flow, join.getFlow());
+    assertEquals("JoinVals", join.getName());
+    assertNotNull(join.leftKeyExtractor);
+    assertNotNull(join.rightKeyExtractor);
+    // FIXME: how to compare?
```
t-novak: How to compare output?
assertEquals does not work because equals is not overidden.

je-ik: `DatasetAssert#unorderedEquals()`

dmvk: i don't think there should be output comparison in this test. That would belong to operator-testkit

euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/Builders.java
```diff
@@ -89,6 +90,22 @@
...
+    /**
+     * Finalizes the operator and retrieves its output dataset.
+     * Using this output new operator {@link MapElements} is added
+     * to the flow to extract values from pairs.
+     *
+     * @return the dataset representing the new operator's output
+     */
+    default Dataset<V> outputValues() {
+      return MapElements
+          .of(this.output())
```
dmvk: shouldn't this operator be named (eg. extract values)?

euphoria-core/src/test/java/cz/seznam/euphoria/core/client/operator/JoinTest.java
dmvk: simple instantce comparison should be enough here (`assertTrue(x == y)`)

PR #219, opened 2017-12-13
Hadoop tuning

dmvk:


euphoria-core/src/main/java/cz/seznam/euphoria/core/util/ExceptionUtils.java
```diff
@@ -0,0 +1,50 @@
...
+      throw new RuntimeException(e);
+    }
+  }
+
+  public interface SupplyingAction<T> {
+    T perform() throws Exception;
+  }
+
+  public interface Action<T> {
+    void perform() throws Exception;
```
je-ik: A more standard name would be `apply`.

euphoria-core/src/main/java/cz/seznam/euphoria/core/util/ExceptionUtils.java
```diff
@@ -0,0 +1,50 @@
...
+      action.perform();
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new RuntimeException(e);
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  public interface SupplyingAction<T> {
```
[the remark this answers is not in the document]
je-ik: Also naming seems a little unclear. Could this be `Factory` or just `Supplier`? For better usability it could be handy if this would extend `Serializable`.

euphoria-hadoop/build.gradle
```diff
@@ -1,7 +1,20 @@
...
+    exclude group: 'org.slf4j'
+  }
+  compileOnly("org.apache.hadoop:hadoop-mapreduce-client-core:${hadoopVersion}") {
+    exclude group: 'log4j'
+    exclude group: 'org.slf4j'
+  }
+  compileOnly("org.apache.hadoop:hadoop-yarn-common:${hadoopVersion}") {
+    exclude group: 'log4j'
+    exclude group: 'org.slf4j'
+  }
```
je-ik: I think you can exclude the dependency globally by something like
```
configurations {
  compile.exclude group: "log4j"
  compile.exclude group: "org.slf4j"
}
```

euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/HadoopUtils.java
```diff
@@ -15,127 +15,45 @@
...
-    TaskAttemptID taskAttemptID = new TaskAttemptID(
-            "0", // job tracker ID
-            0, // job number,
-            TaskType.REDUCE, // task type,
-            taskNumber, // task ID
-            0); // task attempt
-    return new TaskAttemptContextImpl(conf, taskAttemptID);
-  }
+  private static final JobID JOB_ID = new JobID(
+      Long.toString(Clock.systemDefaultZone().millis()),
```
je-ik: Why not `System.currentTimeMillis()`?

euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/output/HadoopSink.java
```diff
@@ -30,47 +33,40 @@
...
-              HadoopUtils.createTaskContext(conf.getWritable(), 0));
-
-      committer.setupJob(HadoopUtils.createJobContext(conf.getWritable()));
-    } catch (Exception ex) {
-      throw new RuntimeException(ex);
-    }
+    ExceptionUtils.unchecked(() -> {
+      final TaskAttemptContext setupContext =
+          HadoopUtils.createSetupTaskContext(conf.get(), jobID.get());
+      newOutputFormatInstance()
```
je-ik: I think this violates hadoop OutputFormat's contract. We must call:
 1) checkOutputSpecs()
 2) getRecordWriter()
 3) getOutputCommitter()

All of this in the *same* instance of the OutputFormat. There can be some logic that relies on this and the contract has to be preserved.

je-ik: This means we have to lazy initialize instance of `OutputFormat` in `transient` field and when the instance is created it has to be configured and setup.

euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/output/HadoopSink.java
```diff
@@ -152,20 +141,21 @@ public void write(Pair<K, V> record) throws IOException {
       try {
         hadoopWriter.write(record.getFirst(), record.getSecond());
       } catch (InterruptedException e) {
-        throw new IOException(e);
+        Thread.currentThread().interrupt();
```
je-ik: I don't think we should both interrupt thread and throw exception.

euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/output/HadoopSink.java
```diff
@@ -152,20 +141,21 @@ public void write(Pair<K, V> record) throws IOException {
...
     }
 
     @Override
     public void commit() throws IOException {
-      // ~ flush pending changes - if any
       try {
+        // flush pending changes - if any
         hadoopWriter.close(ctx);
       } catch (InterruptedException e) {
         Thread.currentThread().interrupt();
```
je-ik: Same here

euphoria-hadoop/src/test/java/cz/seznam/euphoria/hadoop/output/HadoopSinkTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+    DatasetAssert.unorderedEquals(Arrays.asList(
+        Pair.of(new Text("first"), new LongWritable(1L)),
+        Pair.of(new Text("second"), new LongWritable(2L)),
+        Pair.of(new Text("third"), new LongWritable(3L)),
+        Pair.of(new Text("fourth"), new LongWritable(3L))
+    ), output);
+  }
+
+}
```
je-ik: :+1:

euphoria-hadoop/build.gradle
dmvk: wouldn't his also exclude slf4j that we provide?

euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/output/HadoopSink.java
dmvk: no, this is where the original problem initiated from, HadoopSink is created once per jvm (for the given sink) and needs to handle more than one task (therefore it cannot be reused).

dmvk: agreed that checkOutputSpecs should be called during initalization phase (in driver)

dmvk: main issues is getOutputCommiter method in FileOutputFormat
```
  public synchronized 
     OutputCommitter getOutputCommitter(TaskAttemptContext context
                                        ) throws IOException {
    if (committer == null) {
      Path output = getOutputPath(context);
      committer = new FileOutputCommitter(output, context);
    }
    return committer;
  }
}
```

euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/HadoopUtils.java
dmvk: no reason to use Clock here in static context

dmvk: changed

euphoria-local/src/main/java/cz/seznam/euphoria/executor/local/LocalExecutor.java
```diff
@@ -477,7 +477,6 @@ public void setAccumulatorProvider(AccumulatorProvider.Factory factory) {
                 // end of the stream
                 writer.flush();
                 writer.commit();
-                writer.close();
```
je-ik: I'm not sure here it writer's life cycle doesn't require calling `close`, even after `commit`. That should be clarified and documented.

euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/input/DataSourceInputFormat.java
```diff
@@ -189,7 +189,8 @@ public void close() throws IOException {
 
   private void initialize(Configuration conf) throws IOException {
     if (source == null) {
-      String serialized = conf.get(DATA_SOURCE);
+      final String serialized = conf.get(DATA_SOURCE);
+      desiredSplitSize = conf.getLong(DESIRED_SPLIT_SIZE, 268435456L);
```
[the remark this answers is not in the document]
dmvk: 👍

euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/output/HadoopSink.java
```diff
@@ -39,6 +43,11 @@
   private final SerializableWritable<Configuration> conf;
   private final SerializableWritable<JobID> jobID;
 
+  @GuardedBy("lock")
+  private final transient Map<TaskAttemptID, OutputFormat<K, V>> outputFormats = new HashMap<>();
+
+  private final transient Object lock = new Object();
```
je-ik: This has to cause NPE after deserialization and attempt to synchronize on the lock, because it will be `null`, because of being `transient.

PR #223, opened 2017-12-15
Add windowing to hfilesink

je-ik:


euphoria-hbase/src/main/java/cz/seznam/euphoria/hbase/HFileSink.java
```diff
@@ -99,6 +111,27 @@ public Builder withTable(String table) {
       return this;
     }
 
+    /**
+     * Set zookeeper quorum for the connection.
+     * @param quorum the quorum
+     * @return this
+     */
+    public Builder withZookeeperQuorum(String quorum) {
+      updaters.add(j -> j.getConfiguration().set(
```
dmvk: Is there any reason, this is not consistent with HBase(Source|Sink)? In terms of not using `updaters`

euphoria-hbase/src/main/java/cz/seznam/euphoria/hbase/HFileSink.java
```diff
@@ -250,7 +326,7 @@ public boolean prepareDataset(
   private void moveTo(String p, Path outputDir) {
     try {
       Path source = new Path(outputDir, p);
-      FileSystem fs = source.getFileSystem(getConfiguration());
+      FileSystem fs = source.getFileSystem(rawSink.getConfiguration());
```
dmvk: shouldn't this be in `try (...) {}` block?

PR #225, opened 2017-12-18
Testing

dmvk:


euphoria-examples/src/main/java/cz/seznam/euphoria/examples/wordcount/SimpleWordCount.java
```diff
@@ -172,9 +175,9 @@ private static Flow buildFlow(DataSource<String> input, DataSink<String> output)
     // defined function to these values. The result of this user defined
     // function is then emitted to the output along with its corresponding
     // key.
-    Dataset<Pair<String, Long>> counted = ReduceByKey.named("REDUCE")
+    final Dataset<Pair<String, Long>> counted = ReduceByKey.named("REDUCE")
         .of(words)
-        .keyBy(e -> e)
+        .keyBy(String::toLowerCase)
```
je-ik: Why do we lowercase the token here, in previous version it was not lowercase, do we want to change functionality here?

euphoria-examples/src/test/java/cz/seznam/euphoria/examples/wordcount/SimpleWordCountTest.java
```diff
@@ -0,0 +1,95 @@
...
+            "second:4",
+            "third:4",
+            "fourth:4"
+        );
+      }
+    });
+  }
+
+  @Test
+  public void test_firstLetterUppercase() {
```
je-ik: This test would suggest that we really want to lowercase the input, the question is - why is that?

euphoria-testing/src/main/java/cz/seznam/euphoria/testing/AbstractFlowTest.java
```diff
@@ -0,0 +1,56 @@
...
+ * Abstract test class for user's {@link Flow} testing.
+ *
+ * @param <OUT> type of output dataset
+ */
+public abstract class AbstractFlowTest<OUT> {
+
+  /**
+   * This method describes how the final dataset should look like.
+   *
+   * @return expected output
```
je-ik: We should mention here, that the final list is compared without respect to ordering.

euphoria-examples/src/test/java/cz/seznam/euphoria/examples/wordcount/SimpleWordCountTest.java
dmvk: because i wanted to showcase few simple, that are easy to read and this this was the simplest thing that comes in mind

euphoria-testing/src/main/java/cz/seznam/euphoria/testing/AbstractFlowTest.java
dmvk: 👍

euphoria-examples/src/test/java/cz/seznam/euphoria/examples/wordcount/SimpleWordCountTest.java
je-ik: ok :)

PR #230, opened 2017-12-21
[euphoria-core] performance tuning

je-ik:


euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/OptionalMethodBuilder.java
```diff
@@ -28,10 +28,35 @@
...
 
+  /**
+   * Apply given modifications to builder based on condition.
+   * @param cond the condition to evaluate
+   * @param applyTrue modification to apply when {@code cond} evaluates to {@code true}
+   * @param applyFalse modification to apply when {@code cond} evaluates to {@code false}
+   * @return next step builder
+   */
+  @SuppressWarnings("unchecked")
+  default BUILDER applyIf(
```
dmvk: this should be tested ;)

PR #233, opened 2017-12-21
Flink tuning

je-ik:


je-ik:
Merging this as it is fairly trivial and would block other work.

PR #240, opened 2017-12-22
[euphoria-hadoop] InputSplit doesn't have to be Writable

je-ik:
This should be the correct way to handle serialization in modern Hadoop.

### DIFF
--- a/euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/input/HadoopSplit.java
+++ b/euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/input/HadoopSplit.java
@@ -20,10 +20,8 @@ import cz.seznam.euphoria.core.client.io.UnsplittableBoundedSource;
 import cz.seznam.euphoria.core.client.util.Pair;
 import cz.seznam.euphoria.core.util.ExceptionUtils;
 import cz.seznam.euphoria.hadoop.utils.Cloner;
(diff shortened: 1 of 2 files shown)


euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/input/HadoopSplit.java
```diff
@@ -83,33 +83,49 @@ public HadoopSplit(HadoopSource<K, V> source, InputSplit inputSplit) {
       reader.initialize(inputSplit, ctx);
       return new HadoopReader<>(
           reader, source.getKeyClass(), source.getValueClass(), source.getConf());
-    } catch (Exception e) {
+    } catch (InterruptedException e) {
       Thread.currentThread().interrupt();
       throw new IOException("Interrupted while opening reader.");
```
cervebar: I would suggest to add also the root exception to stack trace to notice user it was originally Interrupted Exception so it's not so standard IOE -> throw new IOException("Interrupted while opening reader.", e );

je-ik: :+1:

je-ik: @cervebar Because the branch is already closed, could you please open a separate PR for this? Thanks a lot!

PR #242, opened 2018-01-04
[euphoria-core] ConfigAware util

mareksimunek:
Provides functions with included config. Useful when its needed pass config with UnaryFunction, BinaryFunctor, UnaryFunctor, ReduceFunctor

### DIFF
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/util/ConfigAware.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/util/ConfigAware.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2016-2018 Seznam.cz, a.s.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


euphoria-core/src/main/java/cz/seznam/euphoria/core/client/util/ConfigAware.java
```diff
@@ -0,0 +1,83 @@
...
+  public interface ConfigAwareBinaryFunctor<LEFT, RIGHT, OUT, CONFIG extends Config>
+      extends Serializable {
+    void apply(LEFT left, RIGHT right, Collector<OUT> ctx, CONFIG config);
+  }
+
+  @FunctionalInterface
+  public interface ConfigAwareReduceFunctor<IN, OUT, CONFIG extends Config>
+      extends Serializable {
+    void apply(Stream<IN> left, Collector<OUT> ctx, CONFIG config);
+  }
```
je-ik: I have also doubts about the ordering of methods and classes here. In the project overall the ordering is different and we should stick to that or change the ordering in global (the first option seems much easier to me :))

dmvk:
obsolete

PR #244, opened 2018-01-11
[euphoria-hadoop] SequenceFileSink builder

mareksimunek:
Breaking  change because constructor of `SequenceFileSink` is now private. Consider to which version of euphoria to put it.

### DIFF
--- a/euphoria-examples/src/main/java/cz/seznam/euphoria/examples/Utils.java
+++ b/euphoria-examples/src/main/java/cz/seznam/euphoria/examples/Utils.java
@@ -140,9 +140,10 @@ class Utils {
       case "hdfs":
       case "file":
         return DataSinks.mapping(
-            new SequenceFileSink<>(
-                ImmutableBytesWritable.class, ImmutableBytesWritable.class,
-                output.toString(), conf),
+            SequenceFileSink.newBuilder(ImmutableBytesWritable.class,  ImmutableBytesWritable.class)
+              .setOutputPath(output.toString())
+              .setConfiguration(conf)
+              .build(),
             b -> Pair.of(new ImmutableBytesWritable(), new ImmutableBytesWritable(b)));
       case "kafka":
         return DataSinks.mapping(
--- a/euphoria-examples/src/main/java/cz/seznam/euphoria/examples/Utils.java
+++ b/euphoria-examples/src/main/java/cz/seznam/euphoria/examples/Utils.java
@@ -24,6 +24,12 @@ import cz.seznam.euphoria.hbase.HBaseSource;
 import cz.seznam.euphoria.kafka.KafkaSink;
 import cz.seznam.euphoria.shadow.com.google.common.base.Preconditions;
 import cz.seznam.euphoria.shadow.com.google.common.base.Strings;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.Cell;
+import org.apache.hadoop.hbase.KeyValue;
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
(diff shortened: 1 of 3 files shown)


euphoria-examples/src/main/java/cz/seznam/euphoria/examples/Utils.java
```diff
@@ -140,9 +140,10 @@ static HBaseSource getHBaseSource(URI input, Configuration conf) {
       case "hdfs":
       case "file":
         return DataSinks.mapping(
-            new SequenceFileSink<>(
-                ImmutableBytesWritable.class, ImmutableBytesWritable.class,
-                output.toString(), conf),
+            SequenceFileSink.newBuilder(ImmutableBytesWritable.class,  ImmutableBytesWritable.class)
```
je-ik: Two spaces, replace by single space, please.

euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/output/SequenceFileSink.java
```diff
@@ -29,21 +33,81 @@
...
-   * @param path the destination where to place the output to
+   * @param keyType the class representing the type of the keys emitted
+   * @param valueType the class representing the type of the values emitted
+   * @param <K> type of the keys emitted
+   * @param <V> type of the values emitted
+   * @return Builder
    */
-  public SequenceFileSink(Class<K> keyType, Class<V> valueType, String path) {
-    this(keyType, valueType, path, new Configuration());
+  public static <K, V> Builder<K, V> newBuilder(Class<K> keyType, Class<V> valueType) {
```
je-ik: This looks imbalanced, if we use `Builder`, I'd suggest to create the builder as empty and set all parameters via methods, e.g. `withKey` and `withValue`, or `withKeyValue`. This way these two parameters have some "special" handling. This would be understandable if it was the case, that *all* required arguments would be set in `newBuilder` method, but output path isn't and therefore it would be better to handle all parameters equally.

dmvk:
I think we squeeze  put this into 0.8 release, although it is a breaking change.

euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/output/SequenceFileSink.java
mareksimunek: I put mandatory fields in newBuilder as param, because otherwise you will need to set explicit type for key and value via `SequenceFileSink.<Text,Text>newBuilder()..` and thats very ugly.

mareksimunek:
I marked constructors as deprecated so it can be in 0.8 release and we can make them private it in next version.

euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/output/SequenceFileSink.java
je-ik: I think that is not entirely true, because the `newBuilder` can return untyped `Builder` object, which can then gather type by additional call to e.g. `withKeyValue` and return correctly typed `Builder` object. You can have an inspiration about how to do this in builders for operators - i.e. `FlatMap`.

mareksimunek:
I redesigned builder API as we discussed to be same as in another classes and to force user to set mandatory fields 
```
SequenceFileSink
        .of(KeyClass, ValueClass)
        .outputPath(outputDir)
        (.withCompression( CompressionClass, CompressionType) )
        (.withConfiguration(conf) )
        .build();
```

In brackets are optional setters.

dmvk:
good job! 👍

PR #251, opened 2018-01-19
[euphoria-hbase] make sure hfiles have correct permissions before bul…

dmvk:
…k loading.

### DIFF
--- a/euphoria-hbase/src/main/java/cz/seznam/euphoria/hbase/HFileSink.java
+++ b/euphoria-hbase/src/main/java/cz/seznam/euphoria/hbase/HFileSink.java
@@ -30,6 +30,7 @@ import cz.seznam.euphoria.core.client.operator.OptionalMethodBuilder;
 import cz.seznam.euphoria.core.client.operator.ReduceByKey;
 import cz.seznam.euphoria.core.client.operator.ReduceWindow;
 import cz.seznam.euphoria.core.client.util.Pair;
+import cz.seznam.euphoria.core.util.ExceptionUtils;
 import cz.seznam.euphoria.hadoop.output.HadoopSink;
 import cz.seznam.euphoria.hadoop.output.HadoopSink.HadoopWriter;
 import cz.seznam.euphoria.shadow.com.google.common.base.Strings;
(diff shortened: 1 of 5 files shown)


euphoria-hbase/src/test/java/cz/seznam/euphoria/hbase/HFileSinkTest.java
```diff
@@ -64,29 +65,15 @@ void loadIncrementalHFiles(Path path) {
...
-    loadedPaths = new ArrayList<>();
-  }
-
-  @After
-  @Override
-  public void tearDown() throws IOException {
-    super.tearDown();
-  }
+  private final List<String> loadedPaths = new ArrayList<>();
+  private final Flow flow = Flow.create();
```
je-ik: I think the flow should be created per test.

dmvk: fixed

euphoria-hbase/src/main/java/cz/seznam/euphoria/hbase/HFileSink.java
```diff
@@ -206,38 +225,46 @@ public HFileSink build() {
...
+  private final String tableName;
   private final boolean doBulkLoad;
   private final byte[][] endKeys;
   @Nullable
   private final Windowing<Cell, ?> windowing;
   private final UnaryFunction<Window<?>, String> folderNaming;
   private final HadoopSink<ImmutableBytesWritable, Cell> rawSink;
 
+  private final String hdfsUser;
+  private final String hbaseUser;
```
je-ik: This seems unused. Why do we do `chmod a+rwx` instead of changing ownership of the resulting hfiles?

cervebar: changed... for now the copied files are the correct user (hbase) and the temporary directorie have RWX so the calling user can delete them... this was problem of previous implementation,that there was permission error after chown

euphoria-hbase/src/main/java/cz/seznam/euphoria/hbase/HFileSink.java
```diff
@@ -390,6 +419,30 @@ void loadIncrementalHFiles(Path path) {
     }
   }
 
+  /**
+   * For the bulk load the files Hbase user (typically hbase) has to have ALL permissions to operate
+   * over HFiles (for move operation). The running user then has to have permissions to delete remaining folders.
+   *
+   * @param conf
+   * @param path
```
je-ik: This will trigger javadoc warning, can we add description of the two parameters or remove the `@param` at all?

PR #257, opened 2018-01-26
[euphoria-flink] flink register classes with serializers

mareksimunek:


euphoria-flink/src/main/java/cz/seznam/euphoria/flink/ExecutionEnvironment.java
```diff
@@ -141,20 +140,42 @@ public static Mode determineMode(Flow flow) {
...
+    toRegister.forEach((Class<?> key, Class<? extends Serializer> value) -> {
+      if (value != null) {
+        environment.registerTypeWithKryoSerializer(key, (Class<? extends Serializer<?>>) value);
+      } else {
+        environment.registerType(key);
+      }
+    });
+  }
+
+  private void registerClassForMode(HashMap<Class<?>, Class<? extends Serializer>> toRegister,
```
dmvk: forMode?

dmvk:
I have only one concern, that class registration is not reusable between spark and flink this way. The best approch would be to find an abstraction that could be used for both, but from my research it is not an easy task. So 👍 for now, but we should create an issue describing the problem, so it could be fixed in the future.

euphoria-flink/src/main/java/cz/seznam/euphoria/flink/ExecutionEnvironment.java
mareksimunek: renamed to `registerClasses`

PR #265, opened 2018-02-12
Slechta/issue 262

slechta:
Fixes a bug where empty partitions and/or leader in kafka were not considered as error.

### DIFF
--- a/euphoria-kafka/src/main/java/cz/seznam/euphoria/kafka/KafkaSource.java
+++ b/euphoria-kafka/src/main/java/cz/seznam/euphoria/kafka/KafkaSource.java
@@ -226,15 +226,18 @@ public class KafkaSource
 
       final long stopAtStamp = stopReadingAtStamp;
       final long defaultOffsetTimestamp = offsetTimestamp;
-      return ps.stream().map(p ->
-          // ~ FIXME a leader might not be available (check p.leader().id() == -1)
-          // ... fail in this situation
-          new KafkaPartition(
-              brokerList, topicId, p.partition(),
-              config,
-              offs.getOrDefault(p.partition(), defaultOffsetTimestamp),
-              stopAtStamp))
-          .collect(Collectors.toList());
+      return ps.stream().map((PartitionInfo p) -> {
+        if (p.leader().id() == -1) {
+          throw new IllegalStateException("Leader not available");
+        }
+
+        return new KafkaPartition(
+          brokerList, topicId, p.partition(),
+          config,
+          offs.getOrDefault(p.partition(), defaultOffsetTimestamp),
+          stopAtStamp);
+        })
+      .collect(Collectors.toList());
     }
   }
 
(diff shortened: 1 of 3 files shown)


euphoria-kafka/src/main/java/cz/seznam/euphoria/kafka/KafkaSource.java
```diff
@@ -219,17 +219,25 @@ public KafkaSource(
...
+          throw new IllegalStateException("Leader not available");
+        }
+
+        return new KafkaPartition(
+          brokerList, topicId, p.partition(),
+          config,
+          offs.getOrDefault(p.partition(), defaultOffsetTimestamp),
+          stopAtStamp);
+        })
+      .collect(Collectors.toList());
```
je-ik: This should be probably indented. Thanks! :-)

slechta:
In order to make the test I would need to mock static method KafkaUtils.newConsumer(). Mockito does not support mocking static methods, therefore I would need adding PowerMock to test dependecies.

dmvk:
Shouldn't we rather replace it with a factory interface, than introducing new dependency?

je-ik:
I don't think it is necessary. Mocking static methods sounds weird to me, I'd suggest the following small refactor:
```java
    @Override
    public UnboundedReader<Pair<byte[], byte[]>, Long> openReader() throws IOException {
      Consumer<byte[], byte[]> c = newConsumer(brokerList, config);
      TopicPartition tp = new TopicPartition(topicId, partition);
      ArrayList<TopicPartition> partitionList = Lists.newArrayList(tp);
      c.assign(partitionList);
      if (startOffset > 0) {
        c.seek(tp, startOffset);
      } else if (startOffset == 0) {
        c.seekToBeginning(partitionList);
      }
      return new ConsumerReader(c, tp, stopReadingAtStamp);
    }

    @VisibleForTesting
    KafkaConsumer<byte[], byte[]> newConsumer(String brokerList, Settings config) {
      return KafkaUtils.newConsumer(brokerList, null, config);
    }
```
which will allow you to override the broker creation in test.

slechta:
That is not sufficient
Mockito cannot mock/spy because :
 - final class

I will move the inner classes out, because it is already a long class. 
(Anyway I am not a big fan of mocking)

euphoria-kafka/src/main/java/cz/seznam/euphoria/kafka/KafkaSource.java
```diff
@@ -206,7 +207,7 @@ public KafkaSource(
             stopReadingAtStamp);
       }
     }
-    try (Consumer<?, ?> c = KafkaUtils.newConsumer(
+    try (Consumer<?, ?> c = newConsumer(
```
dmvk: shouldnt there be `Consumer<byte[], byte[]>`?

je-ik:
Cool, thanks! :+1: Merging.

PR #270, opened 2018-02-23
WIP: Dejv/tez runtime

dmvk:


euphoria-tez/src/main/java/cz/seznam/euphoria/tez/ListCollector.java
```diff
@@ -0,0 +1,81 @@
...
+import cz.seznam.euphoria.core.client.io.Collector;
+import cz.seznam.euphoria.core.client.io.Context;
+import cz.seznam.euphoria.core.util.Settings;
+
+import java.util.ArrayList;
+import java.util.List;
+
+class ListCollector<T> implements Collector<T>, Context {
+
+  private final Window<?> window = GlobalWindowing.Window.get();
```
je-ik: This looks odd. Why is this collector bound to global window only? I think it could accept the window in constructor.

euphoria-tez/src/main/java/cz/seznam/euphoria/tez/ReduceByKeyTranslator.java
```diff
@@ -0,0 +1,71 @@
...
+    final TezAccumulatorProvider accumulatorProvider = TezAccumulatorProvider.of(context);
+
+    if (operator.getWindowing() != null) {
+      final Windowing<IN, W> windowing = operator.getWindowing();
+      inputChain.modify((stream) ->
+          stream.flatMap((element) -> {
+            final Iterable<W> windows =
+                windowing.assignWindowsToElement((TezElement<W, IN>) element);
+            return StreamSupport.stream(windows.spliterator(), false)
+                .map(window -> window);
```
je-ik: Why is this zero-op map needed?

euphoria-tez/src/main/java/cz/seznam/euphoria/tez/FlatMapTranslator.java
```diff
@@ -0,0 +1,74 @@
...
+      this.collector = new ListCollector<>(accumulatorProvider.create());
+
+    }
+
+    @Override
+    public Stream<OUT> apply(IN in) {
+      if (eventTimeExtractor == null) {
+        mapper.apply(in, collector);
+      } else {
+        throw new UnsupportedOperationException("Not implemented.");
```
je-ik: If we use `TezElement` for passing data in operators, it should be just a few lines to implement it here, or am I missing something?

dmvk:
This was not the right direction :)

PR #272, opened 2018-03-06
[euphoria-spark] #243 RDD are cached when used more than once in flow

mareksimunek:
Used different approach than #243. No hints are needed, because it can be deduced from `DAG`, when node has more than one children.

### DIFF
--- a/euphoria-spark/src/main/java/cz/seznam/euphoria/spark/SparkFlowTranslator.java
+++ b/euphoria-spark/src/main/java/cz/seznam/euphoria/spark/SparkFlowTranslator.java
@@ -15,8 +15,6 @@
  */
 package cz.seznam.euphoria.spark;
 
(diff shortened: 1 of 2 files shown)


euphoria-spark/src/main/java/cz/seznam/euphoria/spark/SparkFlowTranslator.java
```diff
@@ -107,8 +105,13 @@
           break;
         }
       }
+
       if (firstMatch != null) {
         final JavaRDD<?> out = firstMatch.translator.translate(op, executorContext);
+        // ~ output result will be used more than once, cache Dataset for reusing
+        if (dag.getNode(op).getChildren().size() > 1) {
+          out.cache();
```
dmvk: wow, this looks like a really simple and clean solution 👍

PR #274, opened 2018-03-12
broadcastHashJoin fix

mareksimunek:
`BroadcastHashJoin` translator was used even when wrong side had `SizeHint.FITS_IN_MEMORY`. (E.g.  in `LeftJoin` only right side with hint should translate to `BroadcastHashJoin` and vice versa)

### DIFF
--- a/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/hint/Util.java
+++ b/euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/hint/Util.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2016-2018 Seznam.cz, a.s.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


euphoria-core/src/main/java/cz/seznam/euphoria/core/client/operator/hint/Util.java
```diff
@@ -0,0 +1,46 @@
...
+package cz.seznam.euphoria.core.client.operator.hint;
+
+import cz.seznam.euphoria.core.client.dataset.Dataset;
+import cz.seznam.euphoria.core.client.dataset.windowing.MergingWindowing;
+import cz.seznam.euphoria.core.client.operator.Join;
+import cz.seznam.euphoria.core.client.operator.Operator;
+
+import java.util.ArrayList;
+
+public class Util {
```
je-ik: The location of this class seems strange. First of all, it seems that it is executor related, so should be in `executor` package, annotated with `Audience` as appropriate. Moreover, I think that the generic name `Util` could be made more explicit.

je-ik:
:+1:

dmvk:
LGTM, good job 👍

Issue #275, opened 2018-04-04
Running with LocalExecutor


djKooks:
Hello,
I'm first in this project, and trying on sample(with gradle).

I'm going with LocalExecutor, but it keep showing 'cannot find LocalExecutor symbol'.
This is my gradle file.
```
dependencies {
    // This dependency is found on compile classpath of this component and consumers.
    ...
    compile 'cz.seznam.euphoria:euphoria-core:0.7.0'
    compile 'cz.seznam.euphoria:euphoria-local:0.5.0'
    ...
```
It seems it cannot find `euphoria-local 0.5.0` package. Could you let me know correct package name for this?
One more, if I want to use `Spark` or `Flink` executor, is there a guide for that?

Thanks.

je-ik:
Hi, please use the same version for all `euphoria-*` artifacts. The name `LocalExecutor` (and module `euphoria-local`) was first introduced in release `0.8.0-rc1`, so using this version should fix your problem. In versions 0.7.0 ane before, the module was named `euphoria-inmem` and the executor was `InMemExecutor`. Please let us know if this fixed your issue.

Regarding Spark or Flink executor, you can use one of them by adding `euphoria-spark` or `euphoria-flink` artifacts and using `SparkExecutor` or `FlinkExecutor`. There should be some examples in the `examples` module. Documentation is currently a work-in-progress. :-)

djKooks:
@je-ik Thanks for quick response!

One more, can be a dumb question...
when I want to use Spark/Flink Executor, do I have to install Spark/Flink in local first?

je-ik:
@djKooks that depends ... if you want to use spark/flink for tests only, then this should not be needed, because the dependencies will be downloaded from maven central. That means you would instantiate the executor as follows for instance:
```
  SparkExecutor executor = SparkExecutor.newBuilder("test").local().build();
```
That would give you executor that you can use in unit tests. If you want to test your flow on real executor, then you would need a running spark or flink cluster (either local or remote). You then submit your flow using `spark-submit` or `flink` commands.

djKooks:
@je-ik Thanks for help. But I still have some failure with sample...
```
        String inputPath = "input.txt";
        String outputPath = "output.txt";
        DataSource<String> dataSource = new SimpleHadoopTextFileSource(inputPath);
        DataSink<String> dataSink = new SimpleHadoopTextFileSink<>(outputPath);

        // Define a flow, i.e. a chain of transformations
[...]

PR #276, opened 2018-04-04
Fix code in README.md

je-ik:


je-ik:
README.md was out of sync with the code, as mentioned in #275

djKooks:
@je-ik I just tested with fixed code. Thanks.
One thing more, how about wrap Executor part with try-catch in end of code in README? (it is defined at SimpleWordCount.java, but not here)

```
...
try {
    Executor executor = new LocalExecutor();
    executor.submit(flow).get();
} catch (Exception e) {
    // do exception handling
}
```

je-ik:
@djKooks sorry, I didn't notice your comment. That sounds good, opened #277

PR #279, opened 2018-04-24
[euphoria-spark] Optimize kryo serialization.

dmvk:


euphoria-spark/src/main/java/cz/seznam/euphoria/spark/kryo/SingletonSerializer.java
```diff
@@ -0,0 +1,79 @@
...
+
+  /** Cached singleton instance. */
+  private T instance;
+
+  private SingletonSerializer(String methodName) {
+    this.methodName = methodName;
+  }
+
+  @Override
+  public void write(Kryo kryo, Output output, T object) {}
```
je-ik: Probably better indentation with some comment about why is this method empty?

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/kryo/SingletonSerializer.java
```diff
@@ -0,0 +1,79 @@
...
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public T read(Kryo kryo, Input input, Class<T> type) {
+    try {
+      if (instance == null) {
```
je-ik: Do we need to guarantee that the method will be invoked only once? Maybe not, but if so we should probably make some explicit synchronization here.

dmvk: I'm pretty sure we don't, but I'll rather synchronize it anyway.

PR #282, opened 2018-04-26
[euphoria-beam] wip

je-ik:


VaclavPlajt:
The Travis build was restarted without changes. And it was surprisingly enough to make the build successful.

PR #288, opened 2018-05-09
[euphoria-spark] Add batch join implementation.

dmvk:


euphoria-spark/src/main/java/cz/seznam/euphoria/spark/JoinIterator.java
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Given an iterator that emits a sequence of tuples as a result of some full-outer join, filters
+ * out those tuples that don't conform to a given join-type.
+ *
+ * <p>In other words, the following are removed:
+ *
+ * <p>* no entries for a full-outer join, * entries with an empty left-side value for a left join
+ * and * entries with an empty right-side value for a right join.
+ */
+class JoinIterator<K, L, R> extends AbstractIterator<Tuple2<K, Tuple2<Optional<L>, Optional<R>>>> {
```
VaclavPlajt: Could we have a bit more fitting name. e.g. `JoinIteratorFilteringByJoinType`?

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/JoinIterator.java
```diff
@@ -0,0 +1,75 @@
...
+            return item;
+          }
+          break;
+        case RIGHT:
+          if (value._2.isPresent()) {
+            return item;
+          }
+          break;
+        default:
+          throw new AssertionError("Unexpected Join.Type [" + joinType + "].");
```
VaclavPlajt: Using Predicate selected/created in constructor to determine which items should be filtered can be another way of implementing this. We can also have exception to be thrown during construction.

PR #289, opened 2018-05-09
[euphoria-spark] Optimize reduce by key.

dmvk:


euphoria-spark/src/main/java/cz/seznam/euphoria/spark/KeyedWindow.java
```diff
@@ -66,4 +68,17 @@ public String toString() {
             ", key=" + key +
             '}';
   }
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public int compareTo(KeyedWindow<W, K> o) {
+    Preconditions.checkArgument(
+        o.key instanceof Comparable,
+        "Key is not comparable, this is probably a bug in euphoria-spark translation.");
```
je-ik: I think this check should be removed in final implementation, maybe so FIXME here?

dmvk: I think it should stay there, as it is illegal state, which means KeyedWindow was used the way it was not intended.

je-ik: I'm generally reserved to testing preconditions in methods of this kind, because they are called **very** often and the check either fails every time or not at all. The performance penalty might be negligible, but it creates sort of "code spam". Misuse of the object should be IMHO covered by tests.

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/ReduceByKeyTranslator.java
```diff
@@ -93,17 +104,46 @@ static boolean wantTranslate(ReduceByKey operator) {
...
+      // we can n
+      reduced = tuples.groupByKey().flatMapValues(new Reducer<>(reducer, accumulatorProvider));
     }
+
+    return reduced.map(
+        t -> {
+          final KeyedWindow<W, KEY> kw = t._1();
+          final OUT el = t._2();
+          return new SparkElement<>(kw.window(), kw.timestamp(), Pair.of(kw.key(), el));
+        });
```
VaclavPlajt: The `doTranslate()` method has grown to more than 100 lines and there are several nested structures. It hurts readability all together. New appropriately named methods may help.

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/ReduceByKeyIterator.java
```diff
@@ -0,0 +1,76 @@
...
+      if (currentKey.equals(previousKey)) {
+        // inner iterator did not consume all values for a given key, we need to skip ahead until we
+        // find value for the next key
+        inner.next();
+        continue;
+      }
+
+      previousKey = currentKey;
+
+      return new Tuple2<>(
```
VaclavPlajt: Returned `Iterable` will have no items in it when outer `ReduceByKeyIterator.next()` will be called. It would be nice do mention this fact in docs.

PR #294, opened 2018-06-08
optional setting of LazyOutputFormat for HadoopSink

davidak09:
`LazyOutputFormat` doesn't create empty files (`part-m-XXXX`, `part-r-XXXX`)

### DIFF
--- a/euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/output/HadoopSink.java
+++ b/euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/output/HadoopSink.java
@@ -30,6 +30,7 @@ import org.apache.hadoop.mapreduce.OutputFormat;
 import org.apache.hadoop.mapreduce.RecordWriter;
 import org.apache.hadoop.mapreduce.TaskAttemptContext;
 import org.apache.hadoop.mapreduce.TaskAttemptID;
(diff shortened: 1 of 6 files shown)


davidak09:
OK, I agree that tests are really 'must have' here. So I tried to parameterize  existing `HadoopSinkTest` and added tests with `LazyOutputFormat` turned on/off for each of `SequenceFileSink`, `HadoopTextFileSink`, `SimpleHadoopTextFileSink` and `HadoopToStringSink`

Can you please look at it? Should I squash commits?

je-ik:
The commit history looks OK to me.

davidak09:
thanks! 

I had to add another small improvement because the way `SequenceFileSink builder` was coded it forced user to call either `withConfiguration` or `withCompression` before he/she could call `withLazyOutputFormat`.

Personally I would prefer simple builder pattern to this complicated builder cascade which on one hand forces user to follow some kind of convention when calling builder methods (which is good) but also makes code almost unreadable and way too complex even though it could and should be pretty simple class...

mareksimunek:
Yes cascade builder were discussed several times and it was concluded it needs some different project which will generate cascade builders. Little simplification of this did @VaclavPlajt on 
https://github.com/seznam/beam/pull/6  where you dont need to specify params with each new builder and dont create cyclic dependecy on builders.

PR #295, opened 2018-06-15
SequenceFileSink builder enhancements

davidak09:
I found some solution to the builder problem, so here's an enhancement proposal. It looks more readable, at least to me. :-)

Inspiration taken from: https://blog.jayway.com/2012/02/07/builder-pattern-with-a-twist/

### DIFF
--- a/euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/output/SequenceFileSink.java
+++ b/euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/output/SequenceFileSink.java
@@ -51,11 +51,11 @@ public class SequenceFileSink<K, V> extends HadoopSink<K, V> {
    * @param <V> type of the values emitted
    * @return OutputPathBuilder
    */
-  public static <K, V> OutputPathBuildable<K, V> of(Class<K> keyClass, Class<V> valueClass) {
+  public static <K, V> OutputPathBuilder<K, V> of(Class<K> keyClass, Class<V> valueClass) {
     return new Builder<>(keyClass, valueClass);
   }
 
-  public static class Builder<K, V> implements OutputPathBuildable<K, V>, Buildable<K, V> {
+  public static class Builder<K, V> implements OutputPathBuilder<K, V>, OptionalBuilder<K, V> {
 
     private final Class<K> keyClass;
     private final Class<V> valueClass;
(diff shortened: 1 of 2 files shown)


dmvk:
Interesting, it seems like a reasonable trade-off between code complexity and usability.

This could be probably also applied to generic type builders, eg.: ReduceByKey

cc @VaclavPlajt 

```
static <IN> KeyBuilder<IN>of(Dataset<IN> input) {
  // ...
}

interface KeyBuilder<IN> {
  <K> ReduceBuilder<IN, K> keyBy(UnaryFunction<IN, K> keyExtractor);
}

interface ReduceBuilder<IN, K> {
  <OUT> FinalBuilder<IN, K, OUT> keyBy(ReduceFunctor<IN, OUT> reducer);
}

interface FinalBuilder<IN, K, OUT> {
  // optinal builders
}

class Builder<IN, K, OUT> extends KeyBuilder<IN>, ReduceBuilder<IN, K>, FinalBuilder<IN, K, OUT> {
}
```

euphoria-hadoop/src/main/java/cz/seznam/euphoria/hadoop/output/SequenceFileSink.java
```diff
@@ -52,263 +51,143 @@
...
-        Class<K> keyClass,
-        Class<V> valueType,
-        String outputPath,
-        boolean useLazyOutputFormat) {
-      this.keyClass = keyClass;
-      this.valueClass = valueType;
-      this.outputPath = outputPath;
-      this.useLazyOutputFormat = useLazyOutputFormat;
-    }
+  public interface Buildable<K, V> {
```
dmvk: Can we have a better name for this interface? For example `OptinalBuilder`? I also think that we should stick with `*Builder` naming convention for interfaces

davidak09: OK, I renamed `Buildable` to `OptionalBuilder` and `OutputPathBuildable` to `OutputPathBuilder`

VaclavPlajt:
That is a clever solution to our builders problem. I elaborated little bit more. And it seem to suit to our needs. See [an example](https://github.com/VaclavPlajt/playground/blob/master/src/main/java/cz/plastique/examples/ClassToBuild.java).

dmvk:
@VaclavPlajt awesome, this should simplify things a lot

PR #296, opened 2018-06-28
LROB-40 `BatchJoinIterator` extended to log some key and element statistics

VaclavPlajt:


euphoria-spark/src/main/java/cz/seznam/euphoria/spark/BatchJoinIterator.java
```diff
@@ -143,4 +165,57 @@ private void emitLeft() {
...
+
+    if(!LOG.isInfoEnabled()){
+      return;
+    }
+
+    LOG.info("-- {} statistics:", BatchJoinIterator.class.getSimpleName());
+    LOG.info("-- Keys count: {}, input elements count: {}, output (emitted/joined) elements #: {}.",
+        numOfElementsByKey.size(), numOfEncounteredElements, numOfEmittedElements);
+
+    List<Entry<K, Tuple2<MutableLong, MutableLong>>> top10Keys = numOfElementsByKey.entrySet()
```
dmvk: this can introduce a memory problem, it basically means that we need to fit all keys for a single task in memory. Can we keep track of top N elements using priority queue?

dmvk: That would mean that we need to fit only N keys in memory

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/BatchJoinIterator.java
```diff
@@ -143,4 +165,57 @@ private void emitLeft() {
...
+    }else {
+      leftRightCounts._2.increment();
+    }
+
+    numOfEncounteredElements++;
+  }
+
+  private void logStats(){
+
+    if(!LOG.isInfoEnabled()){
```
je-ik: Absolutely minor comment - we should be consistent in code style - if is not a function and therefore is followed by space. It just "looks better" :-)

euphoria-spark/src/main/java/cz/seznam/euphoria/spark/BatchJoinIterator.java
VaclavPlajt: I believe that we can keep top N elements only when pre-grupped by key. Or in another words keys coming from wrapped iterators do not change until all elements with the same key from input dataset are depleted. Does this assumption hold? It should when we are doing join, but I was not entirely sure.
I will re implement it soon.

dmvk: yes, all elements always come sorted by KEY and secondary sorted by SIDE (left side first) ;)

PR #304, opened 2018-10-24
[euphoria-beam-io] Added beam wrappers for euphoria Sinks and Sources

mareksimunek:


euphoria-beam-io/src/main/java/cz/seznam/euphoria/beam/io/EuphoriaIO.java
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+  }
+
+  /**
+   * Write transform for euphoria sinks.
+   *
+   * @param <T> type of data to write
+   */
+  public static class Write<T> extends PTransform<PCollection<T>, PDone> {
```
dmvk: we should get rid of Write transform completely

euphoria-beam-io/src/main/java/cz/seznam/euphoria/beam/io/EuphoriaIO.java
```diff
@@ -0,0 +1,234 @@
+package cz.seznam.euphoria.beam.io;
```
dmvk: licsense

mareksimunek: added

mareksimunek:
@dmvk EuphoriIO.writer deleted, added lincenses.

gradle/versions.gradle
```diff
@@ -17,7 +17,7 @@ ext {
   mockitoVersion = '2.10.0'
 
   // ~ benchmark dependencies
-  beamVersion = '0.6.0'
+  beamVersion = '2.7.0'
```
dmvk: Should the version be 2.8.0?

mareksimunek: If 2.8.0 would be released then yes :), but its not yet

PR #307, opened 2019-09-03
[euphoria-kafka] upgrade jackson-core version to prevent security rep…

je-ik:
…orts from github


je-ik:
No so fast. :-)
This is a little schizofrenic, because travis build on gradle. Can maven be completely removed?

